PR #4, opened 2023-06-06
add update-package-lock workflow

dlockhart:
Depends on:
https://github.com/Brightspace/repo-settings/pull/1139


.github/CODEOWNERS
```diff
@@ -1 +1,2 @@
 *       @BrightspaceUI/gaudi-dev
+package-lock.json
```
bearfriend, line 2:
```suggestion
package-lock.json

```

bearfriend (approved):

svanherk (approved):

.github/CODEOWNERS
jameswklassen, line 2: This is maybe a dumb question but what does it mean to have `package-lock.json` in the codeowners file?

dlockhart: It's essentially saying that if a change is happening only to the lock file (which is what this `update-package-lock` workflow does), then there are no CODEOWNERS and our requirement that PRs are approved by a CODEOWNER goes away, enabling auto-merge.

PR #7, opened 2023-06-06
Add full web-test-runner createConfig()

bearfriend:
Adds a method to get a default wtr config object with some shortcut options and the ability to override.

[US152244](https://rally1.rallydev.com/#/?detail=/userstory/699786552875&fdp=true): Visual-diff: create a shared WTR config

The current `core` wtr config would be replicated with the following:
```javascript
import { createConfig, getBrowsers } from '@brightspace-ui/testing';

const pattern = type => `+(components|controllers|directives|helpers|mixins|templates)/**/*.${type}.js`;

export default createConfig({
	pattern,
	groups: [{
		name: 'aXe',
		files: pattern('axe'),
		browsers: getBrowsers(['chromium'])
	}]
});

```

~Note that the `--playwright` and `--browsers` CLI options will no longer work as we force usage of `playwrightLauncher`,~ but any number of `groups` can be added and configured as necessary to be able to run different tests with different browsers from the command line.


src/wtr-config.js
```diff
@@ -0,0 +1,105 @@
...
+	};
+
+	const config = {
+		...defaultConfig,
+		...timeoutConfig,
+		...passthroughConfig
+	};
+
+	if (vdiff) {
+		config.reporters ??= [ defaultReporter() ];
```
bearfriend, line 94: This was sort of a guess as to how this should work. I'll look into whether `defaultReporter()` should always be included or not.

src/wtr-config.js
```diff
@@ -0,0 +1,106 @@
+import { defaultReporter } from '@web/test-runner';
+import { playwrightLauncher } from '@web/test-runner-playwright';
+import { visualDiff } from './visual-diff-plugin.js';
+import { visualDiffReporter } from './visual-diff-reporter.js';
+
+const DEFAULT_PATTERN = type => `./**/*.${type}.js`;
+const DEFAULT_VDIFF = true;
```
dlockhart, line 7: I think we'll want this to be `false` by default since visual diff will still require some configuration within a new repo.

src/wtr-config.js
```diff
@@ -0,0 +1,106 @@
+import { defaultReporter } from '@web/test-runner';
+import { playwrightLauncher } from '@web/test-runner-playwright';
+import { visualDiff } from './visual-diff-plugin.js';
+import { visualDiffReporter } from './visual-diff-reporter.js';
+
+const DEFAULT_PATTERN = type => `./**/*.${type}.js`;
```
dlockhart, line 6: Probably 95% of repos (or more) use `./test/**/*.test.js` as the pattern, just not core.

src/wtr-config.js
```diff
@@ -0,0 +1,106 @@
...
+}
+
+export function createConfig({
+	pattern = DEFAULT_PATTERN,
+	vdiff = DEFAULT_VDIFF,
+	timeout,
+	...passthroughConfig
+} = {}) {
+
+	console.assert(typeof pattern === 'function', 'pattern must be a function');
```
dlockhart, line 62: I haven't seen this pattern before... I don't think it stops execution like a `throw` would? I'd probably just throw `TypeError`s in these scenarios.

src/wtr-config.js
```diff
@@ -0,0 +1,106 @@
+import { defaultReporter } from '@web/test-runner';
+import { playwrightLauncher } from '@web/test-runner-playwright';
+import { visualDiff } from './visual-diff-plugin.js';
+import { visualDiffReporter } from './visual-diff-reporter.js';
+
+const DEFAULT_PATTERN = type => `./**/*.${type}.js`;
+const DEFAULT_VDIFF = true;
+const DEFAULT_BROWSERS = ['chromium', 'firefox', 'webkit'];
+
+export function getBrowsers(browsers = DEFAULT_BROWSERS) {
```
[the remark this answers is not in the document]
bearfriend: The problem is that if we define the browsers manually to get `reducedMotion` we can't use `--browsers` from the command line.
Screen Shot 2023-06-07 at 9 51 06 AM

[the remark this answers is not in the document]
bearfriend: Actually, I have an idea. Let me run this a couple different ways to test.

package.json
```diff
@@ -6,14 +6,17 @@
...
+    "test:server": "mocha ./test/server/**/*.test.js"
   },
   "author": "D2L Corporation",
   "license": "Apache-2.0",
   "devDependencies": {
+    "chai": "^4.3.7",
     "eslint": "^8",
-    "eslint-config-brightspace": "^0.23"
+    "eslint-config-brightspace": "^0.23",
+    "mocha": "^10.2.0"
```
dlockhart, line 19: Nit: typically we'll just go with something like `^10` since the `^` automatically installs minor/patch versions anyway. Unless you require >= 10.2. 🤷

.eslintrc.json
dlockhart, line 1: I'll muck around with the lint configs in my PR. I'll probably need to split them into two -- one for the Node.js stuff and one for the code that'll run in the browser.

src/server/wtr-config.js
```diff
@@ -0,0 +1,124 @@
...
+		};
+	}
+
+	const defaultConfig = {
+		files: pattern('test'),
+		nodeResolve: true,
+		testRunnerHtml: testFramework =>
+			`<html lang="en">
+				<body>
+					<script src="./tools/resize-observer-test-error-handler.js"></script>
```
dlockhart, line 94: This file doesn't exist anymore now that we're outside of core. Since core will ultimately use this library, we _probably_ want to copy it here. Then eventually once _everyone_ has switched to `@brightspace-ui/testing`, we could remove it from core.

src/server/wtr-config.js
```diff
@@ -0,0 +1,124 @@
...
+		...timeoutConfig,
+		...passthroughConfig
+	};
+
+	config.groups ??= [];
+	config.groups.push({
+		name: 'unit',
+		files: pattern('test'),
+		browsers: getBrowsers()
+	});
```
dlockhart, line 111: What's the outcome of this... `web-test-runner --group unit` runs with reduced motion (and would be what we recommend to folks) but `web-test-runner` on its own or with `--group default` runs without?

src/server/wtr-config.js
```diff
@@ -0,0 +1,124 @@
...
+const DEFAULT_VDIFF = false;
+const ALLOWED_BROWSERS = ['chromium', 'firefox', 'webkit'];
+
+const requestedBrowsers = argv.toString().match(new RegExp(ALLOWED_BROWSERS.join('|'), 'g'));
+
+if (argv.includes('default')) {
+	if (argv.includes('--playwright')) {
+		console.warn('Warning: reducedMotion disabled. Use the unit group to enable reducedMotion.');
+	}
+	else {
```
dlockhart, line 15:
```suggestion
	} else {
```

src/server/wtr-config.js
```diff
@@ -0,0 +1,124 @@
+import { argv } from 'node:process';
+import { defaultReporter } from '@web/test-runner';
+import { playwrightLauncher } from '@web/test-runner-playwright';
+
+const DEFAULT_PATTERN = type => `./test/**/*.${type}.js`;
+const DEFAULT_VDIFF = false;
+const ALLOWED_BROWSERS = ['chromium', 'firefox', 'webkit'];
+
+const requestedBrowsers = argv.toString().match(new RegExp(ALLOWED_BROWSERS.join('|'), 'g'));
```
dlockhart, line 9: Nice!

dlockhart (commented):
I like how this is shaping up! Seems like it's in a good spot to flesh out the tests to cover the new code. We can delay README docs a bit until we're more comfortable things aren't changing.

src/server/wtr-config.js
bearfriend: Right, and they'll get warnings that it's running in puppeteer/without reduced motion from [here](https://github.com/BrightspaceUI/testing/pull/7/files#diff-3725d91fd7c5e6547acd4c933f5f7b6cace48cf03d9099f6c31506d116c13f0aR11-R18) (which is currently a bit flawed).

dlockhart (approved):
Looking really good!

src/server/wtr-config.js
```diff
@@ -0,0 +1,157 @@
...
+				</html>`,
+		};
+
+		const config = {
+			...defaultConfig,
+			...timeoutConfig,
+			...passthroughConfig
+		};
+
+		config.groups ??= [];
```
dlockhart, line 107: Had to [look this up](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), cool!

test/server/wtr-config.test.js
```diff
@@ -0,0 +1,120 @@
+import { createConfig, getBrowsers, WTRConfig } from '../../src/server/wtr-config.js';
+import { createConfig as createConfigPublic, getBrowsers as getBrowsersPublic } from '../../src/index.js';
+import { expect } from 'chai';
+import sinon from 'sinon/pkg/sinon-esm.js';
```
dlockhart, line 4: I'd suggest explicitly importing the things you need, and you can just use the base import now.

```suggestion
import { assert, restore, stub } from 'sinon';
```

test/server/wtr-config.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+describe('createConfig()', () => {
```
dlockhart, line 6: It might make sense to wrap all of these in a single "createWtrConfig" suite. That way as the tests are reported they'll all be grouped under that and it'll be easy to tell what functionality they belong to if one fails.

A single parent suite also lets you do common cleanup tasks like `sinon.restore`.

test/server/wtr-config.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+	it('calls WTRConfig.create', () => {
+		const createStub = sinon.stub(WTRConfig.prototype, 'create').callsFake(num => num / 2);
+		const res = createConfig(10);
+
+		sinon.assert.calledOnce(createStub);
+		sinon.assert.calledWith(createStub, 10);
+		expect(res).to.equal(5);
+
+		createStub.restore();
```
dlockhart, line 20: I'd suggest moving the `restore` call into a common `afterEach`. Otherwise if one of these `assert`s fails, `restore` won't get called and the rest of your tests will start behaving weirdly, possibly making it more difficult to diagnose what actually went wrong.

test/server/wtr-config.test.js
```diff
@@ -0,0 +1,120 @@
...
+
+	it('should enable nodeResolve', () => {
+		expect(config.nodeResolve).to.be.true;
+	});
+
+	it('should not configure a timeout by default', () => {
+		expect(config.testFramework).to.be.undefined;
+	});
+
+	it('should set timeout when provided with a timeout value', () => {
```
dlockhart, line 47: Not sure how exhaustive we want to get, but you could test some of the exception handling. (e.g. setting the timeout to a non-Number or the pattern to a non-function throws)

src/server/wtr-config.js
```diff
@@ -0,0 +1,157 @@
...
+export class WTRConfig {
+
+	static #singleton;
+	#cliArgs = [];
+
+	constructor() {
+		if (!this.constructor.#singleton) {
+			this.cliArgs = argv;
+			this.constructor.#singleton = this;
+		}
```
dlockhart, line 18: I'm not totally sure what singleton stuff is doing... is it letting you default `cliArgs` to `argv`? Why not just take the CLI args as a parameter to the constructor?

bearfriend: I started writing up an explanation of why I switched away from `constructor(cliArgs)` and realized it didn't work because I'm running my tests against the public `getBrowsers` instead of `WTRConfig.getBrowsers`. Fixing that now.

The singleton part is still useful to avoid unnecessarily creating new instances on every `getBrowsers` call. Not a huge hit to lose it, though.

.eslintrc.json
```diff
@@ -2,5 +2,8 @@
   "extends": "brightspace/node-config",
   "parserOptions": {
     "sourceType": "module"
+  },
+  "env": {
+    "mocha": true
```
svanherk, line 7: Feels like this should live in an `.eslintrc.json` in `test`, but I assume this is what @dlockhart said he'd play with in his PR - not sure where that comment went

package-lock.json
```diff
@@ -14,8 +14,11 @@
         "@web/test-runner-playwright": "^0.10"
       },
       "devDependencies": {
+        "chai": "^4.3.7",
         "eslint": "^8",
-        "eslint-config-brightspace": "^0.23"
+        "eslint-config-brightspace": "^0.23",
+        "mocha": "^10.2.0",
+        "sinon": "^15.1.0"
```
svanherk, line 21: Might need to regenerate this. I _think_ the `update-package-lock` would just fix it, but also wouldn't be surprised if it complained loudly that things were out of sync

package.json
```diff
@@ -6,14 +6,18 @@
   "repository": "https://github.com/BrightspaceUI/testing.git",
   "scripts": {
     "lint": "eslint . --ext .js",
-    "test": "npm run lint && npm run test:headless",
-    "test:headless": "web-test-runner --files \"./test/**/*.test.js\" --node-resolve --playwright"
+    "test": "npm run lint && npm run test:server",
+    "test:headless": "web-test-runner --files \"./test/browser/**/*.test.js\" --node-resolve --playwright",
+    "test:server": "mocha ./test/server/**/*.test.js"
```
svanherk, line 11: Oh also, we still need to update `ci.yml` so these tests are actually running

package.json
```diff
@@ -6,14 +6,18 @@
   "repository": "https://github.com/BrightspaceUI/testing.git",
   "scripts": {
     "lint": "eslint . --ext .js",
-    "test": "npm run lint && npm run test:headless",
-    "test:headless": "web-test-runner --files \"./test/**/*.test.js\" --node-resolve --playwright"
+    "test": "npm run lint && npm run test:server",
```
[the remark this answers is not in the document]
bearfriend: Yeah I removed it after creating the "server" tests. Figured we can add it back when we have them.

src/server/wtr-config.js
```diff
@@ -0,0 +1,157 @@
...
+			};
+		}
+
+		const defaultConfig = {
+			files: pattern('test'),
+			nodeResolve: true,
+			testRunnerHtml: testFramework =>
+				`<html lang="en">
+					<body>
+						<script src="./tools/resize-observer-test-error-handler.js"></script>
```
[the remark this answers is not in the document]
bearfriend: Done locally

package.json
bearfriend: That's done [here](https://github.com/BrightspaceUI/core/pull/3711), if you want to review that one too.

svanherk: But don't we want to run the tests you've written here in this repo's CI?

.github/workflows/ci.yml
```diff
@@ -12,6 +12,10 @@ jobs:
...
           cache: 'npm'
       - name: Install dependencies
-        run: npm ci
+        run: |
+          npm ci
+          npx playwright install-deps
       - name: Lint (JavaScript)
         run: npm run lint
+      - name: Unit Tests
+        run: npm run test
```
svanherk, line 21:
```suggestion
      - name: Lint (Javascript) and Unit Tests
        run: npm run test
```

```suggestion
      - name: Lint (JavaScript)
        run: npm run lint
      - name: Unit Tests (Server)
        run: npm run test:server
```
 Since `npm run test` includes linting, maybe we just remove that step?  Or if we want to keep them separate, we can add separate steps for the server tests (and then eventually the browser tests)?

src/server/wtr-config.js
```diff
@@ -0,0 +1,145 @@
...
+								padding: 30px;
+							}
+							body[data-theme="dark"] {
+								background-color: #000000;
+							}
+							body[data-theme="translucent"] {
+								background: repeating-linear-gradient(45deg, #606dbc, #606dbc 10px, #465298 10px, #465298 20px);
+							}
+						</style>
+						<script type="module" src="./components/typography/typography.js"></script>
```
svanherk, line 45: Oh interesting, what to do with this... 🤔

We definitely don't want to maintain a duplicate of this, maybe we do the same thing as @dlockhart's PR and come up with a "minimum" version?

I wonder if we could make `core` a peer dependency to avoid the circular dependency.  But then what does that do when you try to use this _in_ `core` 😬

.eslintrc.json
dlockhart: Yeah I'll sort it out in mine once this merges.

src/server/wtr-config.js
bearfriend: Oof. Good question. Being peer dependency seems more promising. I'll try setting it up that way and see what happens.

dlockhart: Good catch. If that doesn't work, having minimal typography should be pretty doable.

src/server/wtr-config.js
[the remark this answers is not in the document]
dlockhart: In your other PR, can we update core's version to just import this one? Then we're back to 1 copy. :)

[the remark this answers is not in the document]
svanherk: Clever! So I guess we'd temporarily export it? Which, speaking of, we probably need to add `/tools` anyways to https://github.com/BrightspaceUI/testing/blob/main/package.json#L19-L21?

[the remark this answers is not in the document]
svanherk: Or would this have worked without adding it, because our export needs it? I don't really know how `export` works, compared to regular public `files`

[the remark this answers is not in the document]
dlockhart: We definitely need to add it to `files` in `package.json`, or move it under `src`. For exporting... yeah I'm not sure we might have to experiment. Since it's just being included as `<script src>` maybe whatever logic that prevents folks from importing things we don't export is bypassed.

svanherk (approved):

PR #9, opened 2023-06-07
feat: implementing custom fixture

dlockhart:
Implementation of a custom `fixture()` method that extends the `open-wc` version but recursively waits for nested Lit elements to resolve their `updateComplete`s.

This doesn't currently include special handling for components that would like `fixture` to wait even longer (e.g. after an API call is complete), but that can be a future improvement.


src/fixture/fixture.js
```diff
@@ -0,0 +1,64 @@
+import { nextFrame, fixture as wcFixture } from '@open-wc/testing';
+import { reset } from './reset.js';
+
+function getComposedChildren(node) {
```
dlockhart, line 4: The original draft version used this method from `@brightspace-ui/core`. I don't love this but I had to copy it into here since core is going to eventually use this library and we can't have circular dependencies. If anyone has a better idea...

src/fixture/fixture.js
```diff
@@ -0,0 +1,64 @@
...
+
+let awaitedFonts = false;
+async function waitForFonts() {
+	if (awaitedFonts) return Promise.resolve();
+	awaitedFonts = true;
+	return await document.fonts.ready;
+}
+
+export async function fixture(element, opts) {
+	await Promise.all([reset(opts), waitForFonts()]);
```
dlockhart, line 60: Actually I'm going to add some more tests for `reset` and `waitForFonts`.

test/fixture.test.js
```diff
@@ -0,0 +1,104 @@
...
+
+		afterEach(() => {
+			resolves.clear();
+		});
+
+		it('should wait for element at fixture root', async() => {
+			const finishedPromise = fixture(`<${slowElem} id="elem1"></${slowElem}>`)
+				.then((elem) => elem.finished);
+			await waitUntil(() => resolves.has('elem1'));
+			const timeout = setTimeout(() => resolves.get('elem1')(), 50);
```
dlockhart, line 43: I hate having a real delay in these, but this was the only way I could get these tests to actually fail with our custom `fixture` removed. It's really difficult to test if something is waiting long enough when your test has to also wait that long to prove that it worked!

src/fixture/fixture.js
dlockhart: Thanks, I'll at least try simplifying this to only include the stuff we care about.

Agreed that handling Polymer likely doesn't make any sense, but if we did need to that future plan to `await` a `getLoadingComplete()` Promise could work in a pinch.

dlockhart: I was able to simplify this a bit. Because of the way this encounters shadow roots, this function was never called with a shadow root as the root element, so checking for `ELEMENT_NODE` was sufficient.

src/focus.js
dlockhart, line 1: Including this because I need to call it from my tests. I fully expect this library will eventually export a bunch of mouse/keyboard helpers like this... I'm just not ready to commit to their names/location right now so leaving this internal.

package.json
```diff
@@ -19,7 +19,10 @@
     "mocha": "^10",
     "sinon": "^15"
   },
-  "exports": "./src/index.js",
+  "exports": {
+    ".": "./src/browser/index.js",
+    "./wtr-config.js": "./src/server/index.js"
+  },
```
dlockhart, line 25: Had to split the exports since having them all in one file was causing the browser to blow up on Node.js things.

package-lock.json
```diff
@@ -715,6 +716,90 @@
         }
       }
     },
+    "node_modules/@puppeteer/browsers/node_modules/ansi-styles": {
```
svanherk, line 719: Is it weird that all this `puppeteer` stuff is being added? 🤔

package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "lint": "eslint . --ext .js",
     "test": "npm run lint && npm run test:server",
```
svanherk, line 9:
```suggestion
    "test": "npm run lint && npm run test:server && npm run test:browser",
```
Now do we want this command to include everything?

src/browser/reset.js
```diff
@@ -0,0 +1,60 @@
...
+import { nextFrame } from '@open-wc/testing';
+
+const DEFAULT_LANG = 'en',
+	DEFAULT_VIEWPORT_HEIGHT = 800,
+	DEFAULT_VIEWPORT_WIDTH = 800;
+
+let currentLang = undefined,
+	currentRtl = false,
+	currentViewportHeight = 600,
+	currentViewportWidth = 800;
```
svanherk, line 11: Does the `currentViewportHeight` purposefully start off different from the default to trigger an initial `setViewport`?

bearfriend:
`ci.yml` should be updated again either to include `Unit Test (Browser)`, or combine everything.

src/browser/reset.js
dlockhart: Right, should have put a note about this. Even though I can't find anything in the Playwright documentation about it, it looks like the default viewport when you launch it is 800x600. So yes, this is so that when the first test runs it resizes the viewport to 800x800.

Now since the height of the viewport shouldn't really matter at all anymore, maybe we should just change our default height to 600? The only time it would be used would be if we added an API to snapshot the entire page (which we likely will).

package-lock.json
dlockhart: It's a dependency of `@web/test-runner-chrome` which is a dependency of `@web/test-runner`... so not sure why it was ever _not_ here?

src/browser/reset.js
svanherk: Hmm, yeah we could do that, although what we have now saves us if Playwright ever decides to change their value.  My thinking was maybe setting both to `0` was clearer

package-lock.json
svanherk: Huh yeah, that's super weird. We'll see what the `update` job does 😓

src/browser/reset.js
dlockhart: I like that -- done.

package-lock.json
dlockhart: This was the result of me doing a fresh `npm install` with no lock file, so hopefully it does nothing!

test/browser/fixture.test.js
```diff
@@ -0,0 +1,220 @@
...
+			return this.promise;
+		}
+	}
+);
+
+describe('fixture', () => {
+
+	afterEach(() => restore());
+
+	describe('reset', () => {
```
svanherk, line 38: These tests make me super excited to get to use this new `fixture`!

test/browser/fixture.test.js
```diff
@@ -0,0 +1,220 @@
...
+			resolves.clear();
+		});
+
+		it('should wait for element at fixture root', async() => {
+			const finishedPromise = fixture(`<${slowElem} id="elem1"></${slowElem}>`)
+				.then((elem) => elem.finished);
+			await waitUntil(() => resolves.has('elem1'));
+			const timeout = setTimeout(() => resolves.get('elem1')(), 50);
+			const finished = await finishedPromise;
+			clearTimeout(timeout);
```
svanherk, line 141: Why do we need to clear the timeout?

dlockhart: It's not necessary when everything is working correctly, but when things start to go wrong -- like if `finishedPromise` resolves _before_ the code in the timeout executes -- then the test can fail (as it should) but then the code in the timeout executes during the NEXT test.

But explaining this out gave me an idea to clean them up in `afterEach` which should also make it clear why it's needed, so I'll make that change now!

svanherk: Awesome, yeah I like that much better!

svanherk (approved):

test/browser/fixture.test.js
```diff
@@ -0,0 +1,216 @@
...
+			await fixture(html`<p>hello</p>`);
+			expect(document.documentElement.hasAttribute('dir')).to.be.false;
+		});
+
+		it('should use RTL text direction with AR language', async() => {
+			await fixture(html`<p>mrhban</p>`, { lang: 'ar' });
+			expect(document.documentElement.getAttribute('dir')).to.equal('rtl');
+		});
+
+		it('should default language to EN', async() => {
```
bearfriend, line 99: Should we remove the `lang` attribute first since this is effectively doing the same thing as the test below?
```suggestion
		it('should default language to EN', async() => {
			document.documentElement.removeAttribute('lang');
			expect(document.documentElement.getAttribute('lang')).to.be.null;
```

bearfriend (approved):

PR #12, opened 2023-06-09
copy only the parts of resize observer we actually need

dlockhart:
Actually looking at what this code was doing, I realized that we're guaranteed to be running in the WTR environment here, so we only need 3 lines of this.

Just copying this part and moving it inline (faster) seemed best and easiest. The core won't be able to use it though obviously, but this way we can just starve out the old core version and remove it eventually.


svanherk (approved):
Makes sense

PR #16, opened 2023-06-12
Move visual-diff-plugin to testing repo

bearfriend:
Copied from https://github.com/BrightspaceUI/core/pull/3610

### DIFF
--- a/src/server/visual-diff-plugin.js
+++ b/src/server/visual-diff-plugin.js
@@ -1,0 +1,85 @@
+import { access, constants, stat } from 'node:fs/promises';
+import { dirname, join } from 'path';
+import { env } from 'node:process';
+
+const isCI = !!env['CI'];
+const ciDir = isCI ? 'ci' : '';
+const DEFAULT_MARGIN = 10;
+
+async function checkFileExists(fileName) {
+	try {
+		await access(fileName, constants.F_OK);
+		return true;
+	} catch (e) {
+		return false;
+	}
+}
+
+function extractTestPartsFromName(name) {
+	name = name.toLowerCase();
+	const parts = name.split(' ');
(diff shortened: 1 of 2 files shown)


dlockhart (approved):
This is probably sufficient for now and we can add tests once this is actually doing something useful.

src/server/visual-diff-plugin.js
```diff
@@ -0,0 +1,85 @@
...
+				return;
+			}
+			if (session.browser.type !== 'playwright') {
+				throw new Error('Visual-diff is only supported for browser type Playwright');
+			}
+
+			const browser = session.browser.name.toLowerCase();
+			const { dir, newName } = extractTestPartsFromName(payload.name);
+			const goldenFileName = `${join(dirname(session.testFile), 'screenshots', ciDir, 'golden', browser, dir, newName)}.png`;
+			const currentFileName = `${join(dirname(session.testFile), 'screenshots', ciDir, 'current', browser, dir, newName)}.png`;
```
svanherk, line 52: Doesn't need to be figured out now, but based on the crazy names we've seen in the past hitting Windows limits, wondering if we want to make this shorter.  I'm not particularly attached to `screenshots`

svanherk (approved):

src/server/visual-diff-plugin.js
dlockhart: That's a good point -- could be `vdiff`?

dlockhart: I'm good with being consistent everywhere with `vdiff`.

bearfriend: Merging as-is for now, and will address this probably in the golden migration PR.

PR #17, opened 2023-06-12
feat: visual-diff wait for async elements

dlockhart:
Small improvement to the `fixture` method which allows components to define an optional `getLoadingComplete()` method (similar to `getUpdateComplete()`) and `fixture` will wait for it to resolve before proceeding.

### DIFF
--- a/src/browser/fixture.js
+++ b/src/browser/fixture.js
@@ -29,7 +29,7 @@ function getComposedChildren(node) {
 
 }
 
-async function waitForElem(elem) {
+async function waitForElem(elem, opts) {
 
 	const update = elem?.updateComplete;
 	if (typeof update === 'object' && Promise.resolve(update) === update) {
@@ -37,14 +37,21 @@ async function waitForElem(elem) {
 		await nextFrame();
 	}
 
+	if (opts.awaitLoadingComplete && typeof elem?.getLoadingComplete === 'function') {
+		await elem.getLoadingComplete();
+		await nextFrame();
+	}
+
 	const children = getComposedChildren(elem);
-	await Promise.all(children.map(e => waitForElem(e)));
+	await Promise.all(children.map(e => waitForElem(e, opts)));
 
 }
 
 export async function fixture(element, opts) {
+	opts ??= {};
+	opts.awaitLoadingComplete ??= true;
 	await Promise.all([reset(opts), document.fonts.ready]);
 	const elem = await wcFixture(element);
-	await waitForElem(elem);
+	await waitForElem(elem, opts);
 	return elem;
 }
--- a/test/browser/fixture.test.js
+++ b/test/browser/fixture.test.js
@@ -11,22 +11,43 @@ const resolves = new Map();
 
 const slowElem = defineCE(
 	class extends LitElement {
-		static get properties() {
-			return { nested: { type: Boolean } };
-		}
 		constructor() {
 			super();
-			this.nested = false;
 			this.finished = false;
 			this.promise = new Promise((resolve) => resolves.set(this.id, resolve));
 			this.promise.then(() => this.finished = true);
 		}
 		render() {
-			return this.nested ? html`${unsafeHTML(`<${slowElem} id="nested"></${slowElem}>`)}` : nothing;
+			return html`<slot></slot>`;
 		}
 		async getUpdateComplete() {
 			await super.getUpdateComplete;
 			return this.promise;
+		}
+	}
+);
+
+const asyncElem = defineCE(
+	class extends LitElement {
+		constructor() {
+			super();
+			this.finished = false;
+			this.promise = new Promise((resolve) => resolves.set(this.id, resolve));
+			this.promise.then(() => this.finished = true);
+		}
+		render() {
+			return html`<slot></slot>`;
+		}
+		async getLoadingComplete() {
+			return this.promise;
+		}
+	}
+);
+
+const nestedElem = defineCE(
+	class extends LitElement {
+		render() {
+			return html`${unsafeHTML(`<${slowElem} id="slow"></${slowElem}><${asyncElem} id="async"></${asyncElem}>`)}`;
 		}
 	}
 );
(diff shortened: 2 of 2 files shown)


src/browser/fixture.js
```diff
@@ -29,22 +29,29 @@ function getComposedChildren(node) {
...
-async function waitForElem(elem) {
+async function waitForElem(elem, opts) {
 
 	const update = elem?.updateComplete;
 	if (typeof update === 'object' && Promise.resolve(update) === update) {
 		await update;
 		await nextFrame();
 	}
 
+	if (opts.awaitLoadingComplete && typeof elem?.getLoadingComplete === 'function') {
```
svanherk, line 40: What are the potential use cases for not wanting to wait for `getLoadingComplete`, if it exists?  To vdiff skeletons?

test/browser/fixture.test.js
```diff
@@ -31,6 +27,31 @@ const slowElem = defineCE(
...
+		constructor() {
+			super();
+			this.finished = false;
+			this.promise = new Promise((resolve) => resolves.set(this.id, resolve));
+			this.promise.then(() => this.finished = true);
+		}
+		render() {
+			return html`<slot></slot>`;
+		}
+		async getLoadingComplete() {
```
svanherk, line 41: This is kinda outside the scope of this PR, but worth bring up because I believe James might be starting on the skeleton group loading stuff soon - do you envision this being the same piece that could be used for the grouped skeletons, to tell the parent "I'm done loading and awaiting confirmation to turn off my skeleton". Or by all HM components in general, to indicate they're done?

Or do you see this as mostly a test/vdiff thing, and those will use cases will look more like the async container stuff?

src/browser/fixture.js
dlockhart: Yeah that was the only use case I could think of, which is why I defaulted this to ON.

test/browser/fixture.test.js
```diff
@@ -31,6 +27,31 @@ const slowElem = defineCE(
...
+			return this.promise;
+		}
+	}
+);
+
+const nestedElem = defineCE(
+	class extends LitElement {
+		render() {
+			return html`${unsafeHTML(`<${slowElem} id="slow"></${slowElem}><${asyncElem} id="async"></${asyncElem}>`)}`;
```
bearfriend, line 50: nit:
```suggestion
			return unsafeHTML(`<${slowElem} id="slow"></${slowElem}><${asyncElem} id="async"></${asyncElem}>`)};
```

test/browser/fixture.test.js
dlockhart: That's a good question, these things are definitely related.

The mechanism as it's used here is more of a 1-time thing that the tests wait for, whereas theoretically the skeleton stuff could go from loading -> ready and then back to loading, in which case the skeleton container and all the items in it should likely go back into skeleton mode.

For that reason my mind went more to the subscriber stuff for the skeletons. This also doesn't need to be set in stone -- once we figure out what works for the skeleton collection we can try this out with it and also try out visual-diff in some hypermedia repos and then adjust as necessary.

src/browser/fixture.js
```diff
@@ -29,22 +29,29 @@ function getComposedChildren(node) {
...
+	}
+
 	const children = getComposedChildren(elem);
-	await Promise.all(children.map(e => waitForElem(e)));
+	await Promise.all(children.map(e => waitForElem(e, opts)));
 
 }
 
 export async function fixture(element, opts) {
+	opts ??= {};
```
bearfriend, line 51: Unless you foresee `null` being passed
```suggestion
export async function fixture(element, opts = {}) {
```

src/browser/fixture.js
```diff
@@ -29,22 +29,29 @@ function getComposedChildren(node) {
...
+	opts.awaitLoadingComplete ??= true;
```
bearfriend, line 52: Just pointing out that this is mutating `opts`, so back in the test suddenly `opts` has this value. Probably not really a big deal here, but my default would be to do:
```javascript
waitForElem(elem, { awaitLoadingComplete: true, ...opts });
```

This also means it doesn't matter was `opts` is, whereas currently `fixure(el, 1)` would fail to set `awaitLoadingComplete`.

test/browser/fixture.test.js
```diff
@@ -31,6 +27,31 @@ const slowElem = defineCE(
 	}
 );
 
+const asyncElem = defineCE(
+	class extends LitElement {
+		constructor() {
+			super();
+			this.finished = false;
+			this.promise = new Promise((resolve) => resolves.set(this.id, resolve));
```
bearfriend, line 35: Taking this opportunity to nit unnecessary parentheses, but if this is a pattern I'm fine with it.
```suggestion
			this.promise = new Promise(resolve => resolves.set(this.id, resolve));
```

dlockhart: Nah that's just an old unnecessary habit of mine -- thanks.

src/browser/fixture.js
dlockhart: I like your suggestion much better, thanks!

test/browser/fixture.test.js
```diff
@@ -31,6 +27,31 @@ const slowElem = defineCE(
...
+		async getLoadingComplete() {
+			return this.promise;
+		}
+	}
+);
+
+const nestedElem = defineCE(
+	class extends LitElement {
+		render() {
+			return unsafeHTML(`<${slowElem} id="slow"></${slowElem}><${asyncElem} id="async"></${asyncElem}>`)};
```
bearfriend, line 50: For shame.
```suggestion
			return unsafeHTML(`<${slowElem} id="slow"></${slowElem}><${asyncElem} id="async"></${asyncElem}>`);
```

src/browser/fixture.js
svanherk: Adding to that... does `waitForElem` need the other opts passed in, or do we think it might eventually? Could this just be
```
await waitForElem(elem, { awaitLoadingComplete: opts.awaitLoadingComplete ?? true });
```
Or simply a boolean parameter if we don't think other options will be needed?

dlockhart: I like the boolean parameter actually -- switched it to that. It's private, so if we need more options later we can always add them.

bearfriend (approved):

PR #18, opened 2023-06-12
Add --grep option to run a subset of test files

bearfriend:
[US152300](https://rally1.rallydev.com/#/?detail=/userstory/699970845705&fdp=true): Visual-diff: run a subset of tests

Allows any library-created `group` to run a subset of test files by modifying the `pattern` with values from the `--grep` (`-g`) CLI arg.

`npm run test:visual-diff -- -g button list` would modify the pattern `./test/**/*.vdiff.js` to `['./test/**/*button*.vdiff.js', './test/**/*list*.vdiff.js']`.

### DIFF
--- a/src/server/wtr-config.js
+++ b/src/server/wtr-config.js
@@ -73,8 +73,23 @@ export class WTRConfig {
 
 	#getPattern(type) {
 		const pattern = this.pattern(type);
-		// if --grep provided, require single-segment wildcards to match
-		return this.#cliArgs.grep?.map(g => pattern.replace(/(?<!\*)\*(?!\*)/g, `*${g}*`)) || pattern;
+
+		// replace filename wildcards with all grep strings
+		// e.g. ./**/test/*.test.* becomes ./**/test/(*grepString*.test.*|*.test.*grepString*)
+		if (this.#cliArgs.grep) {
+			return this.#cliArgs.grep?.map(grepStr => {
+				return pattern.replace(/([^/]*$)/, fileGlob => {
+					const fileGlobs = Array.from(fileGlob.matchAll(/(?<!\*)\*(?!\*)/g)).map(m => {
+						const arr = m.input.split('');
+						arr.splice(m.index, 1, `*${grepStr}*`);
+						return arr.join('');
+					});
+					return `(${fileGlobs.join('|')})`;
+				});
+			});
+		}
+
+		return pattern;
 	}
 
 	create({
--- a/src/server/wtr-config.js
+++ b/src/server/wtr-config.js
@@ -1,6 +1,6 @@
-import { argv } from 'node:process';
 import { defaultReporter } from '@web/test-runner';
 import { playwrightLauncher } from '@web/test-runner-playwright';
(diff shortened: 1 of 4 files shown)


src/server/wtr-config.js
```diff
@@ -1,6 +1,6 @@
-import { argv } from 'node:process';
 import { defaultReporter } from '@web/test-runner';
 import { playwrightLauncher } from '@web/test-runner-playwright';
+import { readCliArgs } from '../../node_modules/@web/test-runner/dist/config/readCliArgs.js';
```
bearfriend, line 3: We likely can use a generic library to get these with less risk.

src/server/wtr-config.js
```diff
@@ -62,15 +62,21 @@ export class WTRConfig {
...
 	create({
 		pattern = DEFAULT_PATTERN,
 		vdiff = DEFAULT_VDIFF,
 		timeout,
 		...passthroughConfig
 	} = {}) {
 
-		if (!this.#cliArgs.includes('--group') || this.#cliArgs.includes('default')) {
-			if (this.#cliArgs.includes('--playwright')) {
+		if (!this.#cliArgs.group?.length || this.#cliArgs.group?.includes('default')) {
```
bearfriend, line 78: Not an array
```suggestion
		if (!this.#cliArgs.group || this.#cliArgs.group === 'default') {
```

src/server/wtr-config.js
```diff
@@ -1,7 +1,16 @@
-import { argv } from 'node:process';
+import commandLineArgs from 'command-line-args';
```
dlockhart, line 1: Will want to add an explicit dependency on this.

bearfriend:
> @eKoopmans, do you remember what syntax you were using with WTR to do this out-of-thebox? I think it was `npx wtr ./**/icon.vdiff.js` but that isn't working for me.

This looks liken it's taking advantage of `--files` being the default option, which, by itself, doesn't work if the group already has a `file` config, and all ours do. In my first commit I used `--files` to do what I'm doing now with `--grep`, but I didn't like mixing the meaning depending on whether you're using a library-defined group or your own group.

eKoopmans:
> @eKoopmans, do you remember what syntax you were using with WTR to do this out-of-thebox? I think it was `npx wtr ./**/icon.vdiff.js` but that isn't working for me.

Okay I haven't looked at this PR so I have no other context, but to answer your question this is the sort of pattern I use:

```
npx web-test-runner **/object-property-list*.test.js --group default
```

test/server/wtr-config.test.js
```diff
@@ -97,6 +97,13 @@ describe('createWtrConfig', () => {
 			expect(config.groups[0]).to.not.have.property('name', 'vdiff');
 		});
 
+		it('should filter test files using --files values', () => {
```
bearfriend, line 100:
```suggestion
		it('should filter test files using --grep values', () => {
```

dlockhart (approved):
Definitely think this covers the use case! 🎉

PR #21, opened 2023-06-14
Add migrate-goldens script

bearfriend:
[US152263](https://rally1.rallydev.com/#/?detail=/userstory/699800157439&fdp=true): Visual-diff: migration for goldens

Usage:
```shell
# POSIX
npx migrate-goldens './test/**'
# Windows
npx migrate-goldens ./test/**
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,1 @@
 node_modules/
-.vdiff/
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 node_modules/
+.vdiff/
--- a/bin/migrate-goldens.js
+++ b/bin/migrate-goldens.js
@@ -1,11 +1,11 @@
 #!/usr/bin/env node
-import { mkdir, rename } from 'node:fs/promises';
 import { join, normalize, sep } from 'path';
-import { glob } from 'glob';
+import { mkdir, rename } from 'node:fs/promises';
 import commandLineArgs from 'command-line-args';
+import { glob } from 'glob';
+import { stdout } from 'node:process';
 
 const { pattern = './test/**' } = commandLineArgs({ name: 'pattern', type: String, defaultOption: true }, { partial: true });
-console.log(pattern);
 const oldSuffix = 'screenshots/ci/golden';
 const newSuffix = 'vdiff/ci/chromium/golden';
 const dirs = await glob(`${pattern}/${oldSuffix}`, { ignore: 'node_modules/**' });
@@ -20,4 +20,4 @@ await Promise.all(dirs.map(async dir => {
 	return rename(dir, join(base, normalize(newSuffix)));
 }));
 
-console.log(`Migrated ${files.length} goldens found in ${dirs.length} test directories`);
+stdout.write(`\nMigrated ${files.length} goldens found in ${dirs.length} test directories\n`);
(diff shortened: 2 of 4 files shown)


bin/migrate-goldens.js
dlockhart, line 1: We can change our mind later, but wondering if we want this script here in this repo, in the old `BrightspaceUI/visual-diff` (which will eventually be graveyarded) or as a separate thing...

bin/migrate-goldens.js
```diff
@@ -0,0 +1,23 @@
...
+import commandLineArgs from 'command-line-args';
+import { glob } from 'glob';
+import { stdout } from 'node:process';
+
+const { pattern = './test/**' } = commandLineArgs({ name: 'pattern', type: String, defaultOption: true }, { partial: true });
+const oldSuffix = 'screenshots/ci/golden';
+const newSuffix = 'vdiff/ci/chromium/golden';
+const dirs = await glob(`${pattern}/${oldSuffix}`, { ignore: 'node_modules/**' });
+const files = [];
+
```
dlockhart, line 13: Could we list out the migration steps somewhere, either in Markdown or in comments? Just so it's super clear which migration steps this handles.

The stuff we need (so far):

1. Rename `screenshots` to `vdiff`
2. Add a new `chromium` directory inside `vdiff\ci\golden`
3. Move everything from the old `screenshots\ci\golden` into `vdiff\ci\golden`
4. "Fix' the directory/filenames to remove redundancies

For 4), I don't know if we can do it without running the tests... because [extractTestPartsFromName](https://github.com/BrightspaceUI/core/pull/3610/files#diff-1efd6ab174aa37274c8e9be049f951ecfd00b2cc15122d8aff895dd11bf3e2f4R19) uses the spaces between suites as important information that we don't have just from the filename because the old library [turned them into hyphens](https://github.com/BrightspaceUI/visual-diff/blob/main/file-helper.js#L44). 🤔 Does it therefore make sense to run this as like a `--migration` flag from the old library?

bin/migrate-goldens.js
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env node
+import { join, normalize, sep } from 'path';
+import { mkdir, rename } from 'node:fs/promises';
+import commandLineArgs from 'command-line-args';
+import { glob } from 'glob';
+import { stdout } from 'node:process';
+
+const { pattern = './test/**' } = commandLineArgs({ name: 'pattern', type: String, defaultOption: true }, { partial: true });
+const oldSuffix = 'screenshots/ci/golden';
```
dlockhart, line 9: For later: I want to revisit all these directory names -- `ci`, `current` & `golden`. But should be easy to adjust this later if we change them.

.gitignore
```diff
@@ -1,2 +1 @@
 node_modules/
-.vdiff/
```
dlockhart, line 2: Did you mean to remove this?

bin/migrate-goldens.js
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env node
+import { appendFile, mkdir, readFile, rename, rm } from 'node:fs/promises';
+import { join, normalize, parse, sep } from 'node:path';
+import commandLineArgs from 'command-line-args';
+import { glob } from 'glob';
+import { stdout } from 'node:process';
+
+const { pattern = './test/**' } = commandLineArgs({ name: 'pattern', type: String, defaultOption: true }, { partial: true });
+const oldSuffix = 'screenshots/ci/golden';
+const newSuffix = 'golden/chromium';
```
dlockhart, line 10: Nit, could reuse from the plugin export:

```suggestion
const newSuffix = `${PATHS.GOLDEN}/chromium`;
```

dlockhart (approved):
Looks good!

dlockhart (approved):

PR #23, opened 2023-06-15
feat: export focusWithKeyboard, focusWithMouse & assert

dlockhart:
Core needs these, so adding them here & exporting. I'll delete [core's copy](https://github.com/BrightspaceUI/core/blob/main/tools/web-test-runner-helpers.js) once it's switched over to use this since it doesn't appear to be used outside of core.

### DIFF
--- a/src/browser/focus.js
+++ b/src/browser/focus.js
@@ -1,6 +1,12 @@
-import { sendKeys } from '@web/test-runner-commands';
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
 
-export const focusWithKeyboard = async(element) => {
-	await sendKeys({ press: 'Escape' });
+export async function focusWithKeyboard(element) {
+	await sendKeys({ press: 'Shift' }); // Tab moves focus, Escape causes dismissible things to close
 	element.focus({ focusVisible: true });
-};
+}
+
+export async function focusWithMouse(element) {
+	const { x, y } = element.getBoundingClientRect();
+	await sendMouse({ type: 'click', position: [Math.ceil(x), Math.ceil(y)] });
+	await sendMouse({ type: 'move', position: [0, 0] });
+}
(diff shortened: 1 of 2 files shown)


src/browser/focus.js
```diff
@@ -1,6 +1,12 @@
-import { sendKeys } from '@web/test-runner-commands';
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
 
-export const focusWithKeyboard = async(element) => {
-	await sendKeys({ press: 'Escape' });
+export async function focusWithKeyboard(element) {
+	await sendKeys({ press: 'Shift' }); // Tab moves focus, Escape causes dismissible things to close
```
dlockhart, line 4: As hinted at with the comment, when using this in core `Escape` was causing our dropdowns and dialogs to dismiss themselves. Core's copy uses `Tab` which was also problematic since it moves focus around and was why our [visual-diff copy](https://github.com/BrightspaceUI/visual-diff/blob/main/helpers/focus.js#L14) used `Escape`.

`Shift` _seems_ to be a safe option... we'll see though I guess.

src/browser/focus.js
```diff
@@ -1,6 +1,12 @@
-import { sendKeys } from '@web/test-runner-commands';
+import { sendKeys, sendMouse } from '@web/test-runner-commands';
 
-export const focusWithKeyboard = async(element) => {
```
svanherk, line 3: Does linting want the `function` setup versus the `const` setup, or just trying to match the style of the repo? Or is there an advantage of one over the other?

svanherk (approved):

src/browser/focus.js
dlockhart: This isn't linting related, it was more of a personal preference. TBH I'm not sure where or why the `const` setup was introduced... _maybe_ [in the old visual-diff library](https://github.com/BrightspaceUI/visual-diff/blob/main/helpers/focus.js#L12) and it got copied a bunch of places? If someone knows more about why we'd want one vs. the other, chime in!

svanherk: I'm not sure, I feel like we use the `const` style randomly throughout in `core`, but it's not consistent. https://github.com/BrightspaceUI/core/blob/main/components/form/form-helper.js
https://github.com/BrightspaceUI/core/blob/main/helpers/gestures.js
And I just flip back and forth depending on what the file is already doing, so was mostly curious!

bearfriend (approved):

PR #24, opened 2023-06-15
Add test debugging setup

bearfriend:
This is just some basic setup to be able to run `--watch` and `--manual`, including automatically pausing in DevTools at the top of each test file.

We can still decide to implement a UI-based play/pause/next system with promises, so I'll link this to [US152298](https://rally1.rallydev.com/#/?detail=/userstory/699970327679&fdp=true) (Visual-diff: ability to debug in the browser), but won't consider it done.


src/browser/reset.js
```diff
@@ -23,7 +23,7 @@ export async function reset(opts) {
 
 	window.scroll(0, 0);
 
-	await sendMouse({ type: 'move', position: [0, 0] });
+	await sendMouse({ type: 'move', position: [0, 0] }).catch(() => {});
```
bearfriend, line 26: In `manual` mode, tests are run in your existing browser installs, which can't be controlled this way, so simply ignoring the failures here, though a console warning may be good. These work fine in `watch` mode.

src/server/wtr-config.js
```diff
@@ -1,13 +1,23 @@
...
 	{ name: 'group', type: String },
-	{ name: 'grep', alias: 'g', type: String, multiple: true }
+	{ name: 'manual', type: Boolean },
+	{ name: 'playwright', type: Boolean },
+	{ name: 'watch', type: Boolean },
+	// custom options
+	{ name: 'chromium', type: Boolean },
+	{ name: 'firefox', type: Boolean },
+	{ name: 'grep', alias: 'g', type: String },
+	{ name: 'filter', alias: 'f', type: String, multiple: true },
```
bearfriend, line 18: `grep` now just forwards to the mocha config and works as it did before. `filter` (open to other names) will do what `grep` did previously: filter test files so you can run your tests in 2 seconds instead of 20.

src/server/wtr-config.js
```diff
@@ -102,52 +152,24 @@ export class WTRConfig {
 		...passthroughConfig
 	} = {}) {
 
-		if (!this.#cliArgs.group || this.#cliArgs.group === 'default') {
-			if (this.#cliArgs.playwright) {
+		const { watch, manual, group, files, playwright	} = this.#cliArgs;
+
+		if (!group || group === 'default') {
+			if (playwright) {
```
dlockhart, line 158: Haha, just wrapping their CLI is looking better and better each time I look at these warnings...

src/server/headed-mode-plugin.js
```diff
@@ -0,0 +1,16 @@
+import { globSync } from 'glob';
```
dlockhart, line 1: Looks like this line is failing CI ^

bearfriend: Yeah, I don't know why. This looks similar to the issue I had last week that Max also [had yesterday](https://d2l.slack.com/archives/C0PHG3QB0/p1686851524778899)

dlockhart (approved):
This is working for me!

PR #25, opened 2023-06-15
US152246 & US152271 - Add basic golden flag behaviour and update directory structure

svanherk:
Ok, this PR is the first in a series of PRs to implement what we've talked about.  This one:
- Adds basic golden flag functionality, passing it to the plugin to use
- Changes to the directory structure we'd talked about (I think lol, let me know if it's wrong)
- adds and exposes chai assertion `.to.be.golden(...)`
- Adds some basic tests, inspired by https://github.com/BrightspaceUI/visual-diff/blob/main/test/test.visual-diff.js

Next PRs will:
1. Clear directories accordingly, depending on the state of `--golden` and `--grep` to match what we talked about in slack
2. Add pixelmatch (for same and different sizes)


package.json
```diff
@@ -8,14 +8,17 @@
     "lint": "eslint . --ext .js",
     "test": "npm run lint && npm run test:server && npm run test:browser",
     "test:browser": "web-test-runner --files \"./test/browser/**/*.test.js\" --node-resolve --playwright",
-    "test:server": "mocha ./test/server/**/*.test.js"
+    "test:server": "mocha ./test/server/**/*.test.js",
+    "test:vdiff": "web-test-runner --config ./test/vdiff/wtr-vdiff.config.js --group vdiff",
```
svanherk, line 12: This can't run in actual CI yet, and won't be able to until we have the action. But for now it's super helpful to run locally to dev the plugin.

src/browser/vdiff.js
```diff
@@ -0,0 +1,10 @@
+import { executeServerCommand } from '@web/test-runner-commands';
+import { expect } from '@open-wc/testing';
+
+export async function screenshotAndCompare(elem, name, opts) {
+	const rect = elem.getBoundingClientRect();
+	const { pass, message } = await executeServerCommand('brightspace-visual-diff', { name, rect, opts });
+	if (!pass) {
+		expect.fail(message);
```
svanherk, line 8: This removes that `expected false to be true` comment, we now just get `AssertionError: No golden exists` or similar

src/server/visual-diff-plugin.js
```diff
@@ -48,8 +60,10 @@ export function visualDiff() {
 
 			const browser = session.browser.name.toLowerCase();
 			const { dir, newName } = extractTestPartsFromName(payload.name);
-			const goldenFileName = `${join(dirname(session.testFile), 'screenshots', ciDir, 'golden', browser, dir, newName)}.png`;
-			const currentFileName = `${join(dirname(session.testFile), 'screenshots', ciDir, 'current', browser, dir, newName)}.png`;
+			const testPath = dirname(session.testFile).replace(rootDir, '');
+			const goldenFileName = `${join(rootDir, '.vdiff', testPath, 'golden', browser, dir, newName)}.png`;
+			const passFileName = `${join(rootDir, '.vdiff', testPath, 'pass', browser, dir, newName)}.png`;
+			const screenshotFileName = `${join(rootDir, '.vdiff', testPath, 'fail', browser, dir, newName)}.png`;
```
svanherk, line 66: This is an example of what this looks like:
image

test/vdiff/vdiff.vdiff.js
```diff
@@ -0,0 +1,88 @@
...
+				}
+			`;
+		}
+		render() {
+			return html`${this.text}<b>Testing</b>`;
+		}
+	}
+);
+
+describe('vdiff-matches', () => {
```
svanherk, line 39: I have these separated to test what multiple folders look like

test/vdiff/vdiff.vdiff.js
```diff
@@ -0,0 +1,88 @@
...
+			elem.text = 'Different Text';
+		} },
+		/*{ name: 'smaller', action: elem => {
+			elem.style.width = '200px';
+			elem.style.height = '50px';
+		} },
+		{ name: 'larger', action: elem => {
+			elem.style.width = '350px';
+			elem.style.height = '70px';
+		} }*/
```
svanherk, line 66: I don't know what I want here yet around testing different sizes, so just leaving this here for now to remind us

test/vdiff/wtr-vdiff.config.js
```diff
@@ -0,0 +1,20 @@
+import { argv } from 'node:process';
+import { createConfig } from '../../src/server/wtr-config.js';
+
+const pattern = type => `test/vdiff/*.${type}.js`;
+
+function getGoldenFlag() {
```
svanherk, line 6: I don't think this will be useful outside this repo, where I need to be able to generate goldens that are different than when a test is running. So I put it in here

test/.eslintrc.json
svanherk, line 1: Or would we rather I move these tests into `browser`?

dlockhart: Since these tests technically run _in_ the browser using WTR I think they should go in there. These are essentially integration tests that test the plugin as well, but in the future if there end up being tests just for the plugin itself (e.g. testing that it moves files correctly or runs pixelmatch correct) then those could go in `server`.

src/server/visual-diff-plugin.js
svanherk: Haha yeah, I can call them something else to make it clearer, although I think this is a good example of the potential garbage pathing you can easily create 😛

src/browser/index.js
```diff
@@ -1,3 +1,4 @@
 export { assert, aTimeout, defineCE, expect, html, nextFrame, oneEvent, waitUntil } from '@open-wc/testing';
 export { focusWithKeyboard, focusWithMouse } from './focus.js';
 export { fixture } from './fixture.js';
+export { screenshotAndCompare } from './vdiff.js';
```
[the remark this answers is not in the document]
svanherk: I can remove this from `index.js` and pull it in directly from `vdiff.js` in the tests for now, if we're not ready to commit. Does it have to be that format?  Like we can't have this?
```
await expect(elem).to.match.golden(opts);
```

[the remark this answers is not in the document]
dbatiste: I kinda like it when it reads like `expect(elem).to.be.visuallyEqual(opts)`, but whatever people like that works! 😄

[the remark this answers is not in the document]
dbatiste: I like that the one Stacey suggested too.

src/server/visual-diff-plugin.js
svanherk: Ok, after meeting with Danny it became clear this was super confusing for no reason lol.  I've updated it, it now looks like this:
image
Still open to debate on whether we like this, but at least it's a bit clearer now what _this_ is.

src/browser/index.js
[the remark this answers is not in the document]
svanherk: Looks like we're limited to any combination of these chain words: https://www.chaijs.com/api/bdd/#language-chains
Plus a custom word at the end, so either of yours would work. Mine does not, would need to be `expect(elem).to.matchGolden(opts);` unless we try to chain and override `match` (probably not worth it).

Any other fun ideas? 😛 
```
await expect(elem).to.be.that.of.golden(opts);
```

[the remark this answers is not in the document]
dbatiste: I think `expect(elem).to.matchGolden(opts);` is pretty good. It's actually more succinct with less chaining, which I like. Easy to remember. The more weird the chaining is, the more likely I'm gonna have to look it up every time. Lol. But, there is also lots of precedent for `.to.be.` so `expect(elem).to.be.golden(opts)` could be nice too.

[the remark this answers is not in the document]
dlockhart: I like `.to.matchGolden()` or `.to.be.golden()`, although neither of them convey that it's a "visual-diff" thing... which may not be a big deal.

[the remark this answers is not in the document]
bearfriend: I like `.to.be.golden()`. Shortest would be `expect(elem).is.golden()` but I'm not sure that fits the `is` pattern, and `.to.be` is much more familiar.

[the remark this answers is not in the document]
svanherk: Ok, added `.to.be.golden(...)`, let me know if you like it!

dlockhart (approved):
Cool that was easy!

PR #29, opened 2023-06-19
feat: basic keyboard and mouse commands

dlockhart:
Introduces some basic keyboard and mouse command wrappers around `@web/test-runner-commands`.

`focusWithKeyboard` has been renamed to just `focus(elem)` for simplicity. I've removed `focusWithMouse` entirely as it's only used a few times in core and since it actually clicks on the element seems a bit too magical. Instead, consumers can just call `click(elem)`.


src/browser/commands.js
```diff
@@ -0,0 +1,39 @@
...
+
+export async function hover(elem) {
+	await sendMouse({ type: 'move', position: getElementPosition(elem) });
+}
+
+export async function hoverAt(x, y) {
+	await sendMouse({ type: 'move', position: [x, y] });
+}
+
+export async function keyboard(action, keys, elem) {
```
dlockhart, line 32: Interested in what folks think about the name here. We can also go with `sendKeys`, but since we've ditch the `@web/test-runner-commands` syntax for the rest it didn't seem necessary to stick with it here.

src/browser/commands.js
```diff
@@ -0,0 +1,39 @@
...
+	if (elem) {
+		await focus(elem);
+	}
+	const val = {};
+	val[action] = keys;
```
dlockhart, line 37: I found the existing API a bit confusing, where you had to call it like `sendKeys({ press: 'A' })` or `sendKeys({ type: '123' })`. This splits out the action (type/press/down/up) from the key as separate parameters.

Open to keeping it the other way around though, it just seemed weird that you could go `sendKeys({ press: 'A', type: 'b' })` and it would explode.

dlockhart: Yikes... trying to use these in a test and I think most of the names are reserved since they exist already on `window` -- i.e. `window.focus()`, `window.hover()`, etc. Will try something else...

bearfriend (approved):

src/browser/index.js
```diff
@@ -1,3 +1,3 @@
 export { assert, aTimeout, defineCE, expect, html, nextFrame, oneEvent, waitUntil } from '@open-wc/testing';
-export { focusWithKeyboard, focusWithMouse } from './focus.js';
+export { clickAt, clickElem, focusElem, hoverAt, hoverElem, sendKeys, sendKeysElem } from './commands.js';
```
svanherk, line 2: I wonder if we should add visual-diff tests for these in here eventually, or if that's overkill because they'll be used in tests all over the consumer repos anyways 🤔

src/browser/commands.js
```diff
@@ -0,0 +1,45 @@
+import { sendKeys as cmdSendKeys, sendMouse } from '@web/test-runner-commands';
+
+function getElementPosition(elem) {
+	const { x, y, width, height } = elem.getBoundingClientRect();
+	return [
+		Math.floor(x + window.scrollX + width / 2),
+		Math.floor(y + window.scrollY + height / 2),
+	];
```
svanherk, line 8: _Super_ minor, but if this was an object with `x` and `y` instead, would make the lines below a tad prettier
```
return clickAt(position.x, position.y);
```

svanherk (approved):
Looks great!

src/browser/index.js
dlockhart: Good call -- I've added unit tests (not visual diff) for the commands, since I'd rather catch bugs in here vs. in consumer code. They're not _super_ useful since they're mostly testing the `test-runner-commands` stuff, but meh.

PR #30, opened 2023-06-19
include doctype to avoid Firefox quirks mode

dlockhart:
While debugging something else in Firefox, I noticed in the console that because we weren't include a DOCTYPE that it was going into "quirks" (aka legacy) mode.


svanherk (approved):
Quirks mode... TIL

PR #32, opened 2023-06-20
US152271 - Cleanup folders in the reporter at the beginning of each run

svanherk:
Option 1 - Cleaning up directories before each run in the reporter. Putting up now for comments, still a bit messy. More comments inline.

This includes the new structure we discussed.

### DIFF
--- a/src/server/visual-diff-reporter.js
+++ b/src/server/visual-diff-reporter.js
@@ -1,0 +1,74 @@
+import { basename, dirname, join } from 'path';
+import { readdirSync, rmSync } from 'node:fs';
+
+function getDiffPaths(dir, _diffPaths) {
+	_diffPaths = _diffPaths || [];
+	const paths = readdirSync(dir, { withFileTypes: true });
+	paths.forEach(path => {
+		const full = join(dir, path.name);
+		const base = basename(path.name);
+
+		if (path.isDirectory()) {
+			if (base === 'pass' || base === 'fail') _diffPaths.push(full);
+			getDiffPaths(full, _diffPaths);
+		} else {
+			if (base === 'report.json') _diffPaths.push(full);
+		}
+	});
+	return _diffPaths;
+}
+
+function cleanupFiles(rootDir, testFiles, updateGoldens, runSubset) {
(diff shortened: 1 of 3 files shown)


src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,74 @@
...
+	const paths = readdirSync(dir, { withFileTypes: true });
+	paths.forEach(path => {
+		const full = join(dir, path.name);
+		const base = basename(path.name);
+
+		if (path.isDirectory()) {
+			if (base === 'pass' || base === 'fail') _diffPaths.push(full);
+			getDiffPaths(full, _diffPaths);
+		} else {
+			if (base === 'report.json') _diffPaths.push(full);
```
svanherk, line 15: Put this in as a placeholder, not sure yet if these are individual html files scattered around, one big file at the base, etc.

src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,74 @@
...
+		}
+	});
+	return _diffPaths;
+}
+
+function cleanupFiles(rootDir, testFiles, updateGoldens, runSubset) {
+	const vdiffDir = join(rootDir, '.vdiff');
+	let pathsToClear = [];
+
```
svanherk, line 24: The logic below is:
- with `--filter` and  `--golden`
  - Delete all directories at the location of the test files to be run (will delete too much if there are sibling and nested test files)
- with `--filter`
  - Delete all `pass` and `fail` directories and `reports` within the location of the test files to be run (will delete too much if there are sibling and nested test files)
- with `--golden`
  - Delete `.vdiff`
- with nothing
  - Delete all `pass` and `fail` directories and `reports` within `.vdiff`

src/server/visual-diff-reporter.js
dlockhart: Probably will be `report.html`?

svanherk: Yeah not sure where I got `json` from lol

svanherk:
Going with https://github.com/BrightspaceUI/testing/pull/36, but can refer back to this if there's issues.

PR #36, opened 2023-06-21
US152271 - Cleanup appropriate folders in the plugin during runs

svanherk:
Option 2 - alternative to https://github.com/BrightspaceUI/testing/pull/32.

After using both I prefer this one - the things that get cleared "make sense" and feel intuitive, whereas the other solution has some compromises.

### DIFF
--- a/src/server/visual-diff-plugin.js
+++ b/src/server/visual-diff-plugin.js
@@ -24,9 +24,11 @@ async function clearDir(updateGoldens, path) {
 	if (updateGoldens) {
 		await rm(path, { force: true, recursive: true });
 	} else {
-		await rm(join(path, PATHS.FAIL), { force: true, recursive: true });
-		await rm(join(path, PATHS.PASS), { force: true, recursive: true });
-		await rm(join(path, 'report.html'), { force: true });
+		await Promise.all([
+			rm(join(path, PATHS.FAIL), { force: true, recursive: true }),
+			rm(join(path, PATHS.PASS), { force: true, recursive: true }),
+			rm(join(path, 'report.html'), { force: true })
+		]);
 	}
 }
 
--- a/src/server/visual-diff-plugin.js
+++ b/src/server/visual-diff-plugin.js
@@ -42,7 +42,7 @@ async function clearDiffPaths(dir) {
 			if (base === 'pass' || base === 'fail') await rm(full, { force: true, recursive: true });
 			else await clearDiffPaths(full);
 		} else {
-			if (base === 'report.html') await rm(full, { force: true, recursive: true });
+			if (base === 'report.html') await rm(full, { force: true });
 		}
 	}
 }
--- a/src/server/visual-diff-plugin.js
+++ b/src/server/visual-diff-plugin.js
@@ -1,6 +1,8 @@
-import { access, constants, mkdir, rename, stat } from 'node:fs/promises';
-import { dirname, join } from 'node:path';
+import { access, constants, mkdir, readdir, rename, rm, stat } from 'node:fs/promises';
+import { basename, dirname, join } from 'node:path';
+import { env } from 'node:process';
 
+const isCI = !!env['CI'];
 const DEFAULT_MARGIN = 10;
 
 async function checkFileExists(fileName) {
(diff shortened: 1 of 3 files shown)


src/server/visual-diff-plugin.js
```diff
@@ -61,9 +101,27 @@ export function visualDiff({ updateGoldens = false } = {}) {
 			const browser = session.browser.name.toLowerCase();
 			const { dir, newName } = extractTestPartsFromName(payload.name);
 			const testPath = dirname(session.testFile).replace(rootDir, '');
-			const goldenFileName = `${join(rootDir, '.vdiff', testPath, 'golden', browser, dir, newName)}.png`;
-			const passFileName = `${join(rootDir, '.vdiff', testPath, 'pass', browser, dir, newName)}.png`;
-			const screenshotFileName = `${join(rootDir, '.vdiff', testPath, 'fail', browser, dir, newName)}.png`;
+			const newPath = join(rootDir, '.vdiff', testPath, dir);
+			const goldenFileName = `${join(newPath, 'golden', browser, newName)}.png`;
+			const passFileName = `${join(newPath, 'pass', browser, newName)}.png`;
+			const screenshotFileName = `${join(newPath, 'fail', browser, newName)}.png`;
```
svanherk, line 107: This switches to the new directory structure we settled on in slack

src/server/visual-diff-plugin.js
```diff
@@ -1,6 +1,8 @@
-import { access, constants, mkdir, rename, stat } from 'node:fs/promises';
-import { dirname, join } from 'node:path';
+import { access, constants, mkdir, readdir, rename, rm, stat } from 'node:fs/promises';
+import { basename, dirname, join } from 'node:path';
+import { env } from 'node:process';
 
+const isCI = !!env['CI'];
```
svanherk, line 5: Added this to make CI as quick as possible, but we could leave it out if we'd rather keep things simple.  It'll just search for `fail`, `pass`, and `reports` that won't exist.

src/server/visual-diff-plugin.js
```diff
@@ -61,9 +102,27 @@ export function visualDiff({ updateGoldens = false } = {}) {
...
+				}
+
+				if (runSubset || currentRun > 0) {
+					if (!clearedDirs.has(newPath)) {
+						const promise = clearDir(updateGoldens, newPath);
+						clearedDirs.set(newPath, promise);
+						await promise;
+					} else {
+						await clearedDirs.get(newPath);
+					}
```
dlockhart, line 123: Nit, slightly less code:
```suggestion
					if (!clearedDirs.has(newPath)) {
						clearedDirs.set(newPath, clearDir(updateGoldens, newPath));
					}
					await clearedDirs.get(newPath);
```

dlockhart (commented):
I agree that this one feels better somehow... like the code responsible for creating the mess is also cleaning it up. Except that the reporter makes the `report.html` files.

svanherk:
> I agree that this one feels better somehow... like the code responsible for creating the mess is also cleaning it up. Except that the reporter makes the report.html files.

And I think we'll need some kind of info manager to pass things between the plugin and the reporter for that, and that manager can own the name for `report.html`

src/server/visual-diff-plugin.js
```diff
@@ -12,6 +14,39 @@ async function checkFileExists(fileName) {
 	}
 }
 
+async function clearDir(updateGoldens, path) {
+	if (updateGoldens) {
+		await rm(path, { force: true, recursive: true });
+	} else {
+		await rm(join(path, 'fail'), { force: true, recursive: true });
+		await rm(join(path, 'pass'), { force: true, recursive: true });
+		await rm(join(path, 'report.html'), { force: true });
```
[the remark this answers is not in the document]
svanherk: Done, but I left `report.html` for now to remind us to handle it - put it in a constant, or the manager, and test that the cleanup happens as expected once it exists

dlockhart (approved):

src/server/visual-diff-plugin.js
```diff
@@ -12,6 +20,39 @@ async function checkFileExists(fileName) {
 	}
 }
 
+async function clearDir(updateGoldens, path) {
+	if (updateGoldens) {
+		await rm(path, { force: true, recursive: true });
+	} else {
+		await rm(join(path, PATHS.FAIL), { force: true, recursive: true });
+		await rm(join(path, PATHS.PASS), { force: true, recursive: true });
+		await rm(join(path, 'report.html'), { force: true });
```
bearfriend, line 29: Can these be done concurrently or does it error out?
```suggestion
		await Promise.all([
			rm(join(path, PATHS.FAIL), { force: true, recursive: true }),
			rm(join(path, PATHS.PASS), { force: true, recursive: true }),
			rm(join(path, 'report.html'), { force: true })
		]);
```

src/server/wtr-config.js
```diff
@@ -188,7 +188,7 @@ export class WTRConfig {
 			//config.reporters.push(visualDiffReporter());
 
 			config.plugins ??= [];
-			config.plugins.push(visualDiff({ updateGoldens: this.#cliArgs.golden }));
+			config.plugins.push(visualDiff({ updateGoldens: golden, runSubset: filter?.length > 0 || grep?.length > 0 }));
```
bearfriend, line 191: I think this ends up achieving the same thing. Could wrap it in `Boolean` if you want.

```suggestion
			config.plugins.push(visualDiff({ updateGoldens: golden, runSubset: filter || grep }));
```

svanherk: Yeah I think I'll wrap it in a boolean, feels weird passing in `null` or the array of values when I don't need to use it at all

src/server/visual-diff-plugin.js
svanherk: Seems to work!

bearfriend (approved):

PR #39, opened 2023-06-22
feat: move elem argument first

dlockhart:
What do folks think about flipping the arguments around here so that `elem` is first? That way it's consistent with `await oneEvent(elem, 'event-name')`.


svanherk (approved):
I like that, we have a lot of other places where the `elem` is first as well

bearfriend:
👍 I went to sleep thinking about this the other night

PR #41, opened 2023-06-24
Recreate golden with this before each test

bearfriend:
Adds a `beforeEach` root hook to mocha, in which the `golden` assertion method is recreated to call `ScreenshotAndCompare` with the current test and fixture element in `this`.

Avoids needing to pass `this` to `golden()`, which also allows arrow functions to be used.

### DIFF
--- a/src/browser/vdiff.js
+++ b/src/browser/vdiff.js
@@ -4,16 +4,16 @@ import { executeServerCommand } from '@web/test-runner-commands';
 mocha.setup({ // eslint-disable-line no-undef
 	rootHooks: {
 		beforeEach() {
-			chai.Assertion.addMethod('golden',
-				(test => function(...args) {
-					return ScreenshotAndCompare.call(test, this._obj, ...args); // eslint-disable-line no-invalid-this
-				})(this.currentTest));
+			const { currentTest } = this;
+			chai.Assertion.addMethod('golden', function(...args) {
+				return ScreenshotAndCompare(currentTest, this._obj, ...args); // eslint-disable-line no-invalid-this
+			});
 		}
 	}
 });
 
-async function ScreenshotAndCompare(elem, opts) {
-	const name = this.fullTitle();
+async function ScreenshotAndCompare(test, elem, opts) {
+	const name = test.fullTitle();
 	const rect = elem.getBoundingClientRect();
 	const { pass, message } = await executeServerCommand('brightspace-visual-diff', { name, rect, opts });
 	if (!pass) {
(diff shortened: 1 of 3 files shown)


src/browser/vdiff.js
```diff
@@ -1,11 +1,20 @@
 import { chai, expect } from '@open-wc/testing';
 import { executeServerCommand } from '@web/test-runner-commands';
 
-chai.Assertion.addMethod('golden', ScreenshotAndCompare);
+mocha.setup({ // eslint-disable-line no-undef
+	rootHooks: {
+		beforeEach() {
+			const { currentTest } = this;
+			chai.Assertion.addMethod('golden', function(...args) {
+				return ScreenshotAndCompare.call({ test: currentTest, elem: this._obj }, ...args); // eslint-disable-line no-invalid-this
```
dlockhart, line 9: I guess `this._obj` is a Chai thing eh, like it's whatever you call `expect(thing).to.be.whatever` with?

src/browser/vdiff.js
```diff
@@ -1,11 +1,20 @@
 import { chai, expect } from '@open-wc/testing';
 import { executeServerCommand } from '@web/test-runner-commands';
 
-chai.Assertion.addMethod('golden', ScreenshotAndCompare);
+mocha.setup({ // eslint-disable-line no-undef
+	rootHooks: {
+		beforeEach() {
+			const { currentTest } = this;
+			chai.Assertion.addMethod('golden', function(...args) {
```
dlockhart, line 8: Any worry about adding the method over and over again? I guess Chai will just overwrite it. Would something like this also work, or does that lose the "this"?

```javascript
let currentTest, obj;

chai.Assertion.addMethod('golden', (...args) => {
	return ScreenshotAndCompare.call({ test: currentTest, elem: obj }, ...args); 
});
mocha.setup({ // eslint-disable-line no-undef
	rootHooks: {
		beforeEach() {
			currentTest = this.currentTest;
			obj = this._obj;
		}
	}
});
```

dlockhart (commented):
Whoa, this is super cool! Makes things so much cleaner.

src/browser/vdiff.js
bearfriend: It seems to just overwrite it, but I agree it feels bad.

Different `this`-es, but this would probably work:
```javascript
let test;

chai.Assertion.addMethod('golden', function(...args) {
	return ScreenshotAndCompare.call({ test, elem: this._obj }, ...args); 
});
mocha.setup({ // eslint-disable-line no-undef
	rootHooks: {
		beforeEach() {
			test = this.currentTest;
		}
	}
});
```

dlockhart (approved):
Amazing!

PR #46, opened 2023-06-28
US152247 & US152272 - Add pixelmatch and handle images of different sizes

svanherk:
Ok, most of the work here was around testing and the resulting code is pretty small. I'll add more comments inline, but at a high-level this PR:
- Adds pixelmatch to compare diffs
  - If the images are the same size, the diff is created and saved and the test fails
  - If the images are different, the screenshot and golden are "grown" accordingly and compared in 9 different positions to find the best diff.  This diff is saved, along with the "grown" images for the report and also for local debugging

Other diffing libraries were investigated but in the end pixelmatch was the simplest and gave us what we needed with minimal code.

### DIFF
--- a/src/browser/vdiff.js
+++ b/src/browser/vdiff.js
@@ -1,15 +1,27 @@
 import { chai, expect } from '@open-wc/testing';
 import { executeServerCommand } from '@web/test-runner-commands';
 
-chai.Assertion.addMethod('golden', ScreenshotAndCompare);
+let test, timeout;
 
-async function ScreenshotAndCompare(testInfo, opts) {
-	const elem = this._obj;
-	const rect = elem.getBoundingClientRect();
-	let results = await executeServerCommand('brightspace-visual-diff-compare', { name: testInfo.test.fullTitle(), rect, opts });
+chai.Assertion.addMethod('golden', function(...args) {
+	return ScreenshotAndCompare.call({ test, timeout, elem: this._obj }, ...args); // eslint-disable-line no-invalid-this
+});
+mocha.setup({ // eslint-disable-line no-undef
+	rootHooks: {
+		beforeEach() {
+			test = this.currentTest;
+			timeout = this.timeout;
+		}
+	}
+});
+
+async function ScreenshotAndCompare(opts) {
+	const name = this.test.fullTitle();
+	const rect = this.elem.getBoundingClientRect();
+	let results = await executeServerCommand('brightspace-visual-diff', { name, rect, opts });
 	if (results.differentSizes) {
-		testInfo.timeout('100000');
-		results = await executeServerCommand('brightspace-visual-diff-compare-different-sizes', { name: testInfo.test.fullTitle() });
+		this.timeout('100000');
+		results = await executeServerCommand('brightspace-visual-diff-compare-different-sizes', { name });
 	}
 	if (!results.pass) {
 		expect.fail(results.message);
--- a/src/browser/vdiff.js
+++ b/src/browser/vdiff.js
@@ -1,16 +1,15 @@
 import { chai, expect } from '@open-wc/testing';
 import { executeServerCommand } from '@web/test-runner-commands';
 
-let test, timeout;
+let test;
 
 chai.Assertion.addMethod('golden', function(...args) {
-	return ScreenshotAndCompare.call({ test, timeout, elem: this._obj }, ...args); // eslint-disable-line no-invalid-this
+	return ScreenshotAndCompare.call({ test, elem: this._obj }, ...args); // eslint-disable-line no-invalid-this
 });
 mocha.setup({ // eslint-disable-line no-undef
 	rootHooks: {
 		beforeEach() {
 			test = this.currentTest;
-			timeout = this.timeout.bind(this);
 		}
 	}
 });
@@ -20,7 +19,7 @@ async function ScreenshotAndCompare(opts) {
 	const rect = this.elem.getBoundingClientRect();
 	let result = await executeServerCommand('brightspace-visual-diff-compare', { name, rect, opts });
 	if (result.resizeRequired) {
-		this.timeout('100000');
+		this.test.timeout('100000');
 		result = await executeServerCommand('brightspace-visual-diff-compare-resize', { name });
 	}
 	if (!result.pass) {
--- a/src/browser/vdiff.js
+++ b/src/browser/vdiff.js
@@ -10,7 +10,7 @@ mocha.setup({ // eslint-disable-line no-undef
 	rootHooks: {
 		beforeEach() {
 			test = this.currentTest;
-			timeout = this.timeout;
+			timeout = this.timeout.bind(this);
 		}
 	}
 });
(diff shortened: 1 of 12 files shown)


src/browser/vdiff.js
```diff
@@ -17,8 +17,12 @@ mocha.setup({ // eslint-disable-line no-undef
 async function ScreenshotAndCompare(opts) {
 	const name = this.test.fullTitle();
 	const rect = this.elem.getBoundingClientRect();
-	const { pass, message } = await executeServerCommand('brightspace-visual-diff', { name, rect, opts });
-	if (!pass) {
-		expect.fail(message);
+	let result = await executeServerCommand('brightspace-visual-diff-compare', { name, rect, opts });
+	if (result.resizeRequired) {
+		this.test.timeout(0);
```
svanherk, line 22:
```suggestion
		this.test.timeout(100000);
```
If we're worried about turning off the timeout, we could also pick a large number. We'll only get to this point after the screenshot is taken, so we don't need to worry about bad tests timing out - those will fail properly.  It'd only be if pixelmatch or the file operations never completed that we'd be stuck.

src/server/visual-diff-plugin.js
```diff
@@ -1,9 +1,12 @@
-import { access, constants, mkdir, readdir, rename, rm, stat } from 'node:fs/promises';
+import { access, constants, mkdir, readdir, readFile, rename, rm, writeFile } from 'node:fs/promises';
 import { basename, dirname, join } from 'node:path';
 import { env } from 'node:process';
+import pixelmatch from 'pixelmatch';
+import { PNG } from 'pngjs';
 
 const isCI = !!env['CI'];
 const DEFAULT_MARGIN = 10;
+const DEFAULT_TOLERANCE = 0; // TODO: Support tolerance override?
```
svanherk, line 9: We currently allow this to be overridden and a few places have, but I don't think we have a story for it

src/server/visual-diff-plugin.js
```diff
@@ -56,6 +59,31 @@ async function clearDiffPaths(dir) {
...
+		{ name: 'center', coord: Math.floor((newSize.height - original.height) / 2) },
+		{ name: 'bottom', coord: newSize.height - original.height }
+	].forEach(y => {
+		[
+			{ name: 'left', coord: 0 },
+			{ name: 'center', coord: Math.floor((newSize.width - original.width) / 2) },
+			{ name: 'right', coord: newSize.width - original.width }
+		].forEach(x => { // TODO: position added for reports, remove/adjust as needed
+			if (original.width === newSize.width && original.height === newSize.height) {
+				resizedPNGs.push({ png: original, position: `${y.name}-${x.name}` });
```
svanherk, line 75: The position can be adjusted to whatever if we want to use it for the report, I don't actually use it anywhere.  I sued it for testing, and because it made these nested `for`s a lot easier to read.

src/server/visual-diff-plugin.js
```diff
@@ -114,56 +139,100 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
-			if (!isCI) { // CI will be a fresh .vdiff folder each time and only one run
-				if (session.testRun !== currentRun) {
-					currentRun = session.testRun;
-					clearedDirs.clear();
-				}
+			const screenshotFile = join(newPath, PATHS.FAIL, browser, newName);
+			const screenshotFileName = `${screenshotFile}.png`;
+
+			if (command === 'brightspace-visual-diff-compare') {
+				if (!isCI) { // CI will be a fresh .vdiff folder each time and only one run
```
svanherk, line 146: I thought about putting this whole cleanup in it's own command to make this cleaner, but then we're running https://github.com/BrightspaceUI/testing/blob/97239155d8cb3a3df32707eea35059c0726d3c13/src/server/visual-diff-plugin.js#L136-L143 3 times per test.  Thoughts?

test/browser/nested/element.vdiff.js
```diff
@@ -0,0 +1 @@
+import '../element.vdiff.js';
```
svanherk, line 1: I've found having this second nested test file was really helpful for all the file manipulation testing, so figured I'd just add it. Could also help with the deleting/creating/rolling up reports testing!

test/browser/element.vdiff.js
```diff
@@ -56,14 +56,24 @@ describe('element-different', () => {
...
 		} },
 		{ name: 'larger', action: elem => {
 			elem.style.width = '350px';
 			elem.style.height = '70px';
-		} }*/
+		} },
+		{ name: 'slimer-taller', action: elem => {
+			elem.style.width = '200px';
+			elem.style.height = '70px';
+			elem.style.textAlign = 'end';
```
svanherk, line 70: Adding these `text-align` changes make these last two tests have interesting best positions chosen

src/browser/vdiff.js
dlockhart: I'm not worried about it -- if we run into these cases we should be able to handle them separately anyway.

src/server/visual-diff-plugin.js
dlockhart: I don't _think_ the reports are going to need it, since they're going to use the resized images anyway which will always both be the same size.

src/server/visual-diff-plugin.js
dlockhart: Yeah... was thinking we could also refactor this a bit at this point to be like:

```javascript
if (command === 'brightspace-visual-diff-compare') {
  doCompare();
} else if (command === 'brightspace-visual-diff-compare-resize') {
  doCompareResize();
}
```

At the very least that'd reduce a lot of the nesting, and would also let us move things out into separate files if that made sense.

src/server/visual-diff-plugin.js
```diff
@@ -114,56 +139,100 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
 
-			const screenshotInfo = await stat(screenshotFileName);
-			const goldenInfo = await stat(goldenFileName);
+				const screenshotImage = PNG.sync.read(await readFile(screenshotFileName));
+				const goldenImage = PNG.sync.read(await readFile(goldenFileName));
+
+				if (screenshotImage.width === goldenImage.width && screenshotImage.height === goldenImage.height) {
+					const diff = new PNG({ width: screenshotImage.width, height: screenshotImage.height });
+					const pixelsDiff = pixelmatch(
+						screenshotImage.data, goldenImage.data, diff.data, screenshotImage.width, screenshotImage.height, { threshold: DEFAULT_TOLERANCE }
```
dlockhart, line 189: Can you pass the `diffMask` option so that they're transparent? I've added it my branch but it keeps getting conflicts/blown away each time I rebase.

dlockhart (approved):
This looks good! I think we'll want to refactor it at some point, but that can wait.

The tricky part is that we need to add extra data for the report before each of the `return` statements, which happen all over the place haha.

src/server/visual-diff-plugin.js
svanherk: Yeah I started doing that and then there were way too many things I was passing around as parameters without knowing what you'd need and where, so I figured that might be better once the skeleton of the report stuff is in

src/server/visual-diff-plugin.js
svanherk: Yeah it was only if we wanted to call it out for extra clarity.  I could remove this and switch to just comments for each object, like
```
	[
		0, // top
		Math.floor((newSize.height - original.height) / 2), // center
		newSize.height - original.height // bottom
	].forEach(y => {
		[
			0, // left
			Math.floor((newSize.width - original.width) / 2), // center
			newSize.width - original.width // right
		].forEach(x => {
```
?

dlockhart: I guess keep it for now and I'll see if it can be used in a useful way. Your TODO can remind us to remove it later if not!

dlockhart (approved):

PR #53, opened 2023-07-04
Basic reporting

dlockhart:
First pass at generating a simple visual-diff HTML report.

Going to be adding a lot more after this, including: filtering, iterating, remembering settings, nicer UI & styles, ability to see all tests for a file, different tabs for the different browsers, and so on.


.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules/
 .vdiff/
+src/server/report/data.js
```
dlockhart, line 3: I don't love that the JSON data goes here, but it was an easy solution to having the files import it and having Rollup "just work". I'm going to be exploring alternatives.

src/server/report/test-result.js
dlockhart, line 1: This component renders out the results for a particular browser for a particular test. They can be stacked or iterated or whatever we want later:

Screenshot 2023-07-04 at 12 23 24 PM

src/server/visual-diff-plugin.js
```diff
@@ -114,6 +115,21 @@ function extractTestPartsFromName(name) {
 	};
 }
```
dlockhart, line 117: This is the code that's saving additional information about each test so that the reporter can ask for it later. As you can see, because there are so many return paths, having to set it from 4 different places isn't ideal. Hopefully some future refactoring can resolve this!

src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,101 @@
...
+		const fileData = files.get(fileName);
+		flattenResults(s, browserData, fileData);
+
+	});
+
+	return { files, browsers };
+
+}
+
+function flattenResults(session, browserData, fileData) {
```
dlockhart, line 39: Test results are nested by "suite", which we don't really care about in the report. This flattens them all (putting " > " in between suites) and converts things into the format that we want.

src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,101 @@
...
+		},
+		stop({ sessions }) {
+
+			if (!reportResults) return;
+
+			const data = createData(rootDir, sessions);
+			const json = JSON.stringify(data, (_key, val) => {
+				if (val instanceof Map) return [...val.values()].sort((a, b) => a.name.localeCompare(b.name));
+				return val;
+			}, '\t');
```
dlockhart, line 93: The resulting data looks like this:

```json
"files": [
	{
		"name": "test/browser/element.vdiff.js",
		"numFailed": 15,
		"tests": [
			{
				"name": "element-different > default",
				"numFailed": 3,
				"results": [
					{
						"name": "Chromium",
						"duration": 81,
						"error": "Image does not match golden. 5061 pixels are different.",
						"passed": false,
						"info": {
							"golden": {
								"path": "test/browser/element-different/golden/chromium/default.png",
								"height": 164,
								"width": 644
							},
							"new": {
								"path": "test/browser/element-different/fail/chromium/default.png",
								"height": 164,
								"width": 644
							},
							"diff": "test/browser/element-different/fail/chromium/default-diff.png"
						}
					}
				]
			}
		]
	}
]
```

```json
"browsers": [
	{
		"name": "Chromium",
		"numFailed": 11,
		"version": "115"
	},
	{
		"name": "Firefox",
		"numFailed": 11,
		"version": "113"
	},
	{
		"name": "Webkit",
		"numFailed": 11,
		"version": "16"
	}
]
```

src/server/visual-diff-plugin.js
```diff
@@ -29,8 +30,7 @@ async function clearDir(updateGoldens, path) {
 	} else {
 		await Promise.all([
 			rm(join(path, PATHS.FAIL), { force: true, recursive: true }),
-			rm(join(path, PATHS.PASS), { force: true, recursive: true }),
-			rm(join(path, 'report.html'), { force: true })
+			rm(join(path, PATHS.PASS), { force: true, recursive: true })
```
dlockhart, line 33: The report stuff is going to handle its own cleanup and will no longer sprinkle `report.html` files everywhere anwyay.

src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,111 @@
...
+
+			const inputDir = join(__dirname, 'report');
+			const reportDir = join(rootDir, PATHS.VDIFF_ROOT, 'report');
+			const tempDir = join(reportDir, 'temp');
+
```
dlockhart, line 98: I didn't like `data.json` being stored in source control, so I've re-jigged this to create `.vdiff/report/temp` where everything gets copied to, Rollup runs from there and dumps stuff into `.vdiff/report/*` and then the `temp` gets removed.

package.json
```diff
@@ -6,6 +6,7 @@
   "repository": "https://github.com/BrightspaceUI/testing.git",
   "scripts": {
     "lint": "eslint . --ext .js",
+    "start": "web-dev-server --root-dir ./.vdiff --open ./report/",
```
svanherk, line 9: When we wrap the runner, can we expose something like this so it doesn't need to be added to every `package.json` for the consumers? Or maybe we run this at the end of a test run 🤔?

dlockhart: I was thinking the same thing but forgot to mention it -- yes! We either do it automatically for visual-diff runs that have more than 1 failure (but not in CI), or add a `--open-report` flag or something (or both).

src/server/report/index.html
```diff
@@ -0,0 +1,21 @@
...
+	<head>
+		<meta http-equiv="Content-Type" content="text/html;charset=utf-8">
+		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<title>Visual Diff Report</title>
+		<link rel="preconnect" href="https://fonts.googleapis.com">
+		<link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
+		<style>
+			@import url('https://fonts.googleapis.com/css2?family=Roboto:wght@400;700&display=swap');
+			body {
+				font-family: 'Roboto', sans-serif;
```
svanherk, line 12: Haha still using a google font but just not Lato? 😛

dlockhart: Haha, didn't want to think about styling anything at this point. We can do Lato, although since we can't actually use any Daylight things from core we're not tied to it.

src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,111 @@
...
+	if (session.testResults) {
+		collectSuite('', session.testResults);
+	}
+
+}
+
+export function visualDiffReporter({ reportResults = true } = {}) {
+	let rootDir;
+	return {
+		start({ config }) {
```
svanherk, line 82: Is there any point in removing the report when you start a run, in case you cancel halfway through and then have a report that doesn't make sense with the current goldens/diffs created?

src/server/visual-diff-reporter.js
```diff
@@ -0,0 +1,111 @@
...
+			rootDir = config.rootDir;
+		},
+		stop({ sessions }) {
```
svanherk, line 85: Does this work with `--watch`? I think I expected this in `onTestRunFinished` or `reportTestFileResults`, but looking at those again they seem to run for each test file which is no good 🤔. But I think ideally in `watch` mode you want a new report after each run

svanherk (commented):
Sorry for the delay in looking at this! Overall setup looks good to me, I didn't worry too much about the details since it's still in flux and we want to do some refactoring to tidy stuff up. Raised a couple concerns

src/server/report/test-result.js
```diff
@@ -0,0 +1,110 @@
...
+			gap: 10px;
+		}
+		.side-by-side > div {
+			flex: 0 1 50%;
+		}
+		img {
+			max-width: 100%;
+		}
+		.diff-container {
+			background: repeating-conic-gradient(#cccccc 0% 25%, #ffffff 0% 50%) 50% / 20px 20px;
```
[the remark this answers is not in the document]
dlockhart: Yeah agreed, and I already have this fixed in the branch I'm working in to add filtering.

I think the behaviour we want is for them to scale to fit the width of the window, but grow no larger than the actual size of the images. That way if you have a small viewport (or a huge image), it'll zoom out to fit everything. I think that works well most of the time, but if it ends up being a problem in certain scenarios we could always have a "show actual size" button or something.

src/server/visual-diff-reporter.js
dlockhart: That's a good call -- done.

src/server/report/test-result.js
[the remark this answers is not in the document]
svanherk: Yeah that sounds good, it's kinda hard to tell in this state with these contrived tests. I suspect we'll still be tweaking the reports as we go through and convert `core`, to see what pieces are actually useful and what's missing, like you did with all the helper commands.

src/server/visual-diff-reporter.js
dlockhart: Ok so I tried this out, and `stop` is not called at all in watch mode. Which kind of jives with their documentation for what it's supposed to do:
> Called once when the test runner stops. This can be used to write a test report to disk for regular test runs.

They call out "regular test runs" which I think implies "not watch runs". I think I'm OK with not writing a report when watching, since you can focus on particular files, go into manual mode and all kinds of other stuff.

svanherk: Yeah, when I was dev-ing in watch mode before reports were a thing I didn't really miss them because you're filtered down to just the tests you're working on and trying to quickly iterate and check out the `resized` or `diff` pngs.

But looking at the docs again, `onTestRunFinished` _might_ give us what we'd need, it does have `sessions`. Does it run only in watch mode and not in regular mode, or would putting the same functionality there end up running twice in regular mode?

dlockhart: `onTestRunFinished` runs in both, and in watch mode it re-runs each time you change a file or focus on a particular file. I actually originally had all this in there, but moved it up into `stop` to avoid having to think about watch mode and reports with partial sessions. I think if it's something we want to support, it's definitely possible.

svanherk: Right, you'd only have the last session that ran passed to `onTestRunFinished`, but data for all the sessions that have been running and you'd probably want to create a full report still to be accurate. Yeah, I don't mind `--watch` clearing the report and not creating a new one, and worrying about supporting it later if we find we want it.

svanherk (approved):

PR #57, opened 2023-07-06
allow filtering on pass/fail and browser

dlockhart:
Introduces a fixed sidebar that contains filters for "status" and "browser". Still just basic styling.

Screenshot 2023-07-06 at 5 45 20 PM


src/server/report/app.js
```diff
@@ -1,35 +1,88 @@
...
+			background-color: #fff;
+			border: 1px solid #e6e6e6;
+			box-shadow: 0 0 6px rgba(0,0,0,.07);
+			box-sizing: border-box;
+			grid-area: sidebar;
+			height: 100vh;
+			position: sticky;
+			top: 0;
+			z-index: 10;
+		}
```
dlockhart, line 37: I stole most of these styles from the Daylight Site. They work for now and we can tweak them later!

src/server/report/app.js
```diff
@@ -1,35 +1,88 @@
...
+			padding-inline: 0;
+		}
+		fieldset > legend {
+			font-weight: bold;
+		}
 	`];
 	constructor() {
 		super();
+		this._filterBrowsers = data.browsers.map(b => b.name);
+		this._filterStatus = data.numFailed > 0 ? FILTER_STATUS.FAILED : FILTER_STATUS.ALL;
```
dlockhart, line 82: Default to showing only failed tests, assuming there are any.

src/server/report/test-result.js
dlockhart, line 1: This fixes a bug Stacey noticed where the images always grow to fill the viewport. Now they'll fill the viewport but not exceed the actual size of the screenshots.

src/server/visual-diff-reporter.js
```diff
@@ -52,7 +63,6 @@ function flattenResults(session, browserData, fileData) {
 			const testData = fileData.tests.get(testName);
 			if (!t.passed) {
 				browserData.numFailed++;
-				fileData.numFailed++;
```
dlockhart, line 55: Counting it here caused it to think that a test failed 3 times if it failed in all 3 browsers, which isn't quite what I want to surface.

src/server/report/app.js
```diff
@@ -1,35 +1,88 @@
...
 		tbody th {
 			font-weight: normal;
 			text-align: left;
 		}
+		td.passed {
+			background-color: #EFFFD9;
+		}
+		td.failed {
+			background-color: #FFEDE8;
+		}
```
svanherk, line 69:
```suggestion
		td.passed {
			background-color: #efffd9;
		}
		td.failed {
			background-color: #ffede8;
		}
```
_Super_ minor lol, and I don't care if we go lowercase or uppercase, but just that they're the same

svanherk (approved):
Looking great!

src/server/report/app.js
dlockhart: For sure -- I was copying and pasting from the Chrome Dev Tools from the Daylight Site and wasn't paying super close attention.

PR #59, opened 2023-07-07
US154765 - Only reset the mouse position when necessary

svanherk:
`sendMouse` can be kinda slow in general - sometimes around 200ms for seemingly no reason.  When a new browser kicks in and starts running this can jump to 1500ms for `webkit` or 2500ms for `firefox`, causing test timeouts and flake.

I investigated playing with the  number of concurrent browsers.  It solves the problem, but I'm seeing differences between CI and local and from repo to repo on whether that makes things faster or slower.

Whereas limiting mouse resets seems to be a benefit overall - it shaves about 40s off local `core` unit test runs.  We'll have to see what it does in CI!

The approach below is fairly simple - it doesn't keep track of where the mouse was moved, just that it _was_ moved, and then will reset it for the next test.

### DIFF
--- a/src/browser/reset.js
+++ b/src/browser/reset.js
@@ -8,7 +8,12 @@ const DEFAULT_LANG = 'en',
 let currentLang = undefined,
 	currentRtl = false,
 	currentViewportHeight = 0,
-	currentViewportWidth = 0;
+	currentViewportWidth = 0,
+	shouldResetMouse = true;
+
+export function requestMouseReset() {
+	shouldResetMouse = true;
+}
 
 export async function reset(opts) {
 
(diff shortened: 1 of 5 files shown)


src/browser/reset.js
```diff
@@ -8,7 +8,12 @@ const DEFAULT_LANG = 'en',
 let currentLang = undefined,
 	currentRtl = false,
 	currentViewportHeight = 0,
-	currentViewportWidth = 0;
+	currentViewportWidth = 0,
+	shouldResetMouse = true;
```
svanherk, line 12: This will cause the mouse to be reset for the first test in every test file. If we think we can count on Playwright to start the mouse at (0, 0) we could default this to `false`, saving maybe an extra 5 seconds overall.

dlockhart: Looking at the code, I think we can assume (0, 0) starting position:
https://github.com/microsoft/playwright/blob/90c92c5724a25e1d9717908581abd35cc0410bde/packages/playwright-core/src/server/input.ts#L165

dlockhart (approved):

src/browser/reset.js
```diff
@@ -23,11 +28,10 @@ export async function reset(opts) {
 
 	window.scroll(0, 0);
 
-	const startTime = Date.now();
-	await sendMouse({ type: 'move', position: [0, 0] }).catch(() => {});
-	const timeTaken = Date.now() - startTime;
-	// eslint-disable-next-line no-console
-	if (timeTaken > 200) console.log(`*** Took ${timeTaken}ms to reset mouse position.`);
+	if (shouldResetMouse) {
```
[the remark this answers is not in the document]
dlockhart: Agreed.

src/browser/reset.js
svanherk: Sounds good, it's also really easy to switch if we learn otherwise

dlockhart (approved):

PR #60, opened 2023-07-07
Rename reports

dlockhart:
This renames the `.vdiff/reports/` directory to `.vdiff/.reports`. It also skips building the report when goldens are being generated, since that would put the report in a weird state.

### DIFF
--- a/src/server/wtr-config.js
+++ b/src/server/wtr-config.js
@@ -185,7 +185,7 @@ export class WTRConfig {
 
 		if (vdiff) {
 			config.reporters ??= [ defaultReporter() ];
-			config.reporters.push(visualDiffReporter());
+			config.reporters.push(visualDiffReporter({ reportResults: !golden }));
 
 			config.plugins ??= [];
 			config.plugins.push(visualDiff({ updateGoldens: golden, runSubset: !!(filter || grep) }));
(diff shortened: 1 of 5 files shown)


src/server/wtr-config.js
```diff
@@ -185,7 +185,7 @@ export class WTRConfig {
 
 		if (vdiff) {
 			config.reporters ??= [ defaultReporter() ];
-			config.reporters.push(visualDiffReporter());
+			config.reporters.push(visualDiffReporter({ reportResults: !golden }));
```
svanherk, line 188: So `reportResults` is something the reporter understands by default? Does this mean that `start` doesn't run and the old report won't be cleaned up?

dlockhart: It's an option that our reporter supports (I just copied the default example). `start` does indeed run and _does_ clean up the old report, but even when I had that being skipped the whole `.vdiff` directory [gets deleted anyway](https://github.com/BrightspaceUI/testing/blob/main/src/server/visual-diff-plugin.js#L40) by the vdiff plugin.

svanherk: Ah ok I see it now, had to expand the whole file.  So yeah, `npm run test:vdiff -- --golden` would've deleted the whole directory, but now `npm run test:vdiff -- --filter nested --golden` will delete the old report and regenerate the goldens for the `nested` tests.

svanherk (approved):

PR #61, opened 2023-07-07
improved bar for test settings

dlockhart:
This introduces a "settings bar" at the top when you're viewing a particular test. It's sticky so that it will float on top in cases where the user needs to scroll vertically.

Screenshot 2023-07-07 at 4 59 08 PM

These styles are still preliminary!

### DIFF
--- a/src/server/report/app.js
+++ b/src/server/report/app.js
@@ -1,20 +1,17 @@
-import './test-result.js';
+import './test.js';
 import { css, html, LitElement, nothing } from 'lit';
+import { FILTER_STATUS, FULL_MODE, LAYOUTS } from './common.js';
+import { classMap } from 'lit/directives/class-map.js';
 import data from './data.js';
 import page from 'page';
-
-const FILTER_STATUS = {
-	ALL: 'All',
-	PASSED: 'Passed',
-	FAILED: 'Failed'
-};
 
 class App extends LitElement {
 	static properties = {
 		_files: { state: true },
 		_filterFile: { state: true },
 		_filterTest: { state: true },
-		_mode: { state: true },
+		_fullMode: { state: true },
+		_layout: { state: true },
 		_overlay: { state: true }
 	};
 	static styles = [css`
@@ -26,7 +23,7 @@ class App extends LitElement {
 		}
 		aside {
 			background-color: #fff;
-			border: 1px solid #e6e6e6;
+			border-right: 1px solid #e6e6e6;
 			box-shadow: 0 0 6px rgba(0,0,0,.07);
 			box-sizing: border-box;
 			grid-area: sidebar;
@@ -41,9 +38,6 @@ class App extends LitElement {
 		main {
 			background-color: #fafafa;
 			grid-area: content;
-		}
-		main > div {
-			padding: 20px;
 		}
 		table {
 			background-color: #ffffff;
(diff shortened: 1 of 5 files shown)


src/server/report/app.js
```diff
@@ -42,9 +39,6 @@ class App extends LitElement {
 			background-color: #fafafa;
 			grid-area: content;
 		}
-		main > div {
-			padding: 20px;
-		}
```
dlockhart, line 47: Because the settings bar needs to fill the full width of the right pane, we can't always add padding.

src/server/report/app.js
```diff
@@ -114,7 +112,13 @@ class App extends LitElement {
 				if (!testData) {
 					view = html`<p>Test not found: <b>${this._filterTest}</b>.</p>`;
 				} else {
-					view = this._renderTest(fileData, testData);
```
dlockhart, line 117: Moved this into its own component.

src/server/report/app.js
```diff
@@ -114,7 +112,13 @@ class App extends LitElement {
...
+					hasPadding = false;
+					view = html`
+						<d2l-vdiff-report-test full-mode="${this._fullMode}" file="${fileData.name}" layout="${this._layout}" ?show-overlay="${this._overlay}" test="${testData.name}" @setting-change="${this._handleSettingChange}"></d2l-vdiff-report-test>
+						<div class="padding">
```
dlockhart, line 118: I think this back button will eventually live in that panel once I add next/previous iterators.

src/server/report/icons.js
dlockhart, line 1: I'm proud of these so be kind! 😆

src/server/report/test.js
```diff
@@ -0,0 +1,159 @@
...
+			border-end-end-radius: 5px;
+			border-inline-end-style: solid;
+			border-start-end-radius: 5px;
+		}
+		.pill input[type="radio"]:checked + label {
+			background-color: #007bff;
+			color: white;
+		}
+		.pill input[type="radio"]:focus-visible + label {
+			z-index: 1;
```
dlockhart, line 83: This was needed otherwise the right side of the outline goes under the next item. If anyone has alternate ideas...

src/server/report/test-result.js
```diff
@@ -37,9 +39,10 @@ class TestResult extends LitElement {
 			position: absolute;
 			top: 0;
 			left: 0;
+			z-index: 1;
```
dlockhart, line 42: This was required so that the sticky settings bar can float overtop of the diff overlay.

src/server/report/app.js
svanherk: If we don't move it to the header, the button seems to jump around a lot as you between full and split mode. I wonder if we could preserve the space so it stays put.

But if we do move it to the header, not an issue!

src/server/report/test.js
```diff
@@ -0,0 +1,159 @@
...
+		test: { type: String },
+	};
+	static styles = [css`
+		.header {
+			background-color: #f0f0f0;
+			border-bottom: 1px solid #e6e6e6;
+			box-shadow: 0 0 6px rgba(0,0,0,.07);
+			position: sticky;
+			top: 0;
+			z-index: 2;
```
svanherk, line 21: Why does this need to be 2 versus just 1?

svanherk (approved):
Looking awesome!

src/server/report/test.js
svanherk: Nvm, I see it now - the overlay is `1`, then the header is `2`

src/server/report/app.js
dlockhart: It's going to the header for sure, but in the next PR where I add tabs for the different tabs it'll jump around a lot less!

src/server/report/test.js
```diff
@@ -0,0 +1,159 @@
...
+
+		return html`
+			<div class="header">
+				<div class="title">
+					<h2>${testData.name}</h2>
+					<div class="file-name">${fileData.name}</div>
+				</div>
+				<div class="settings">
+					${this._renderPillbox('layout', this.layout, [LAYOUTS.FULL, LAYOUTS.SPLIT])}
+					<label class="settings-box"><input type="checkbox" ?checked="${this.showOverlay}" @change="${this._handleOverlayChange}">Overlay Difference</label>
```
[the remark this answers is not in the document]
dlockhart: I know eh... I've left the checkbox styles "browser default" for now, but will likely spend some time on them at some point!

PR #63, opened 2023-07-10
use query-string to drive page.js

dlockhart:
I wasn't quite using the combination of Lit + Page.js + the query-string correctly. This fixes things up.

Essentially:
- The query-string is the source of truth for application state
- When we want to change the state (e.g. apply a filter), we modify the query-string and tell page.js to redirect to that new place
- Page.js the re-runs our init code which parses the query-string and sets the Lit properties from there
- That then triggers a recalculation of `_tests` which is the actual Lit reactive property that renders the application

The end result is that all the filters are on the URL which means you can send people links and the app will "boot up" in the correct state. You can also open the links to individual tests in new tabs and that'll just work as well.

### DIFF
--- a/src/server/report/app.js
+++ b/src/server/report/app.js
@@ -150,13 +150,13 @@ class App extends LitElement {
 		this._updateSearchParams({ file: undefined, test: undefined });
 	}
 	_handleFilterBrowserChange(e) {
-		const browsers = [...this._filterBrowsers];
-		const index = browsers.indexOf(e.target.value);
-		if (!e.target.checked && index > -1) {
-			browsers.splice(index, 1);
-		} else if (e.target.checked && index === -1) {
-			browsers.push(e.target.value);
-		}
+		const browsers = data.browsers.map(b => b.name).filter(b => {
+			if (b === e.target.value) {
+				return e.target.checked;
+			} else {
+				return this._filterBrowsers.includes(b);
+			}
+		});
 		this._updateSearchParams({ browsers: browsers.join(',') });
 	}
 	_handleFilterStatusChange(e) {
--- a/src/server/report/app.js
+++ b/src/server/report/app.js
@@ -80,7 +80,6 @@ class App extends LitElement {
 		this._fullMode = FULL_MODE.GOLDEN.value;
 		this._layout = LAYOUTS.SPLIT.value;
 		this._overlay = true;
-		this._updateFiles();
 	}
 	connectedCallback() {
 		super.connectedCallback();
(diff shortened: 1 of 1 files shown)


src/server/report/app.js
```diff
@@ -80,7 +80,6 @@ class App extends LitElement {
 		this._fullMode = FULL_MODE.GOLDEN.value;
 		this._layout = LAYOUTS.SPLIT.value;
 		this._overlay = true;
-		this._updateFiles();
```
dlockhart, line 83: This isn't needed since Page.js's init code will call this.

src/server/report/app.js
```diff
@@ -141,24 +147,28 @@ class App extends LitElement {
...
-		this._filterStatus = e.target.value;
-		this._updateFiles();
+		this._updateSearchParams({ status: e.target.value });
 	}
 	_handleSettingChange(e) {
 		this[`_${e.detail.name}`] = e.detail.value;
 	}
+	_handleTestClick(e) {
+		this._updateSearchParams({ file: e.target.dataset.file, test: e.target.dataset.test });
+		return false;
```
dlockhart, line 170: Returning `false` from a click handler cases the `href` to not navigate. The `href` is still useful for opening links in new tabs.

src/server/report/app.js
```diff
@@ -247,13 +263,33 @@ class App extends LitElement {
...
+						foundFilterTest = true;
+					}
 				}
 			});
 			if (tests.length > 0) {
 				files.push({ ...f, tests });
 			}
 		});
+
+		if (this._filterTest !== undefined && !foundFilterTest) {
```
dlockhart, line 276: This handles the case where you're viewing a particular test but then apply a filter that will cause that test to be excluded. We zoom back out to app home.

bearfriend (approved):
Getting dangerously close to wanting to implement the proper router, but that can be a later upgrade if we want to go any further.

src/server/report/app.js
```diff
@@ -141,24 +147,28 @@ class App extends LitElement {
...
-			this._filterBrowsers.push(e.target.value);
-		}
-		this._updateFiles();
+		const browsers = data.browsers.map(b => b.name).filter(b => {
+			if (b === e.target.value) {
+				return e.target.checked;
+			} else {
+				return this._filterBrowsers.includes(b);
+			}
+		});
```
dlockhart, line 159: Minor tweak here to use the original browser list to rebuild the filter so that browsers are always rendered in the same order.

PR #65, opened 2023-07-11
Wrap @web/test-runner binary

bearfriend:
[US154135](https://rally1.rallydev.com/#/?detail=/userstory/705070535683&fdp=true): Visual-diff: wrap web-test-runner binary

Use is largely the same. Notable changes:

- Commands: `wtr`/`web-test-runner` -> `d2l-test`
- Browser names: `chromium` -> `chrome`, `webkit` -> `safari`
- Config file:
  - `web-test-runner.config.js` -> `d2l-test.config.js`
  - Now just a simple POJO export. No more `createConfig` or `getBrowsers`
- Options:
  - `--files` is now an explicit option
  - `--group` is now the implicit option, with `unit` being the default (i.e. `d2l-test` and `d2l-test unit` run the `unit` group)
  - `--filter` no longer adds leading and trailing wildcards. So, in core `d2l-test -f list` will _only_ find `list.test.js`. Add your own wildcards to match more: `-f list*` or `-f *list*`

Most wtr options will pass through, but some are explicitly disabled to avoid overwriting the `d2l-test` config.


bin/test.js
```diff
@@ -1,32 +1,44 @@
...
+}) || {};
+
+//const DISALLOWED_ARGS = ['browsers', 'playwright', '_unknown'];
 const DEFAULT_PATTERN = type => `./test/**/*.${type}.js`;
-const DEFAULT_VDIFF = false;
-const ALLOWED_BROWSERS = ['chromium', 'firefox', 'webkit'];
+const ALLOWED_BROWSERS = ['chrome', 'firefox', 'safari'];
+const BROWSER_MAP = { chrome: 'chromium', firefox: 'firefox', safari: 'webkit' };
 
 export class WTRConfig {
```
bearfriend, line 43: Pull this out and leave procedural stuff here.

bin/test.js
```diff
@@ -126,44 +138,34 @@ export class WTRConfig {
 	}
 
 	#getPattern(type) {
-		const pattern = this.#cliArgs.files || this.pattern(type);
+		const pattern = [].concat(this.#cliArgs.files || this.pattern(type));
```
bearfriend, line 141: Allows `--files` option (which will be an array) to be filtered properly

bin/test.js
```diff
@@ -126,44 +138,34 @@ export class WTRConfig {
...
 					const fileGlobs = Array.from(fileGlob.matchAll(/(?<!\*)\*(?!\*)/g)).map(({ index }) => {
 						const arr = fileGlob.split('');
-						arr.splice(index, 1, `*${filterStr}*`);
+						arr.splice(index, 1, filterStr);
 						return arr.join('');
 					});
-					return `+(${fileGlobs.join('|')})`;
-				});
-			});
+					return `+(${fileGlobs.join('|') || fileGlob})`;
```
bearfriend, line 156: Handles bad POSIX CLI globs gracefully.

`--files ./*.test.js` is converted by the OS to a list of all matching files (with no * characters) before being sent to the process.
`--files './*.test.js'` is what _should_ be used. We can add standard POSIX glob note to docs.

bin/test.js
```diff
@@ -126,44 +138,34 @@ export class WTRConfig {
...
-
-		const { files, filter, golden, grep, group, manual, playwright, watch } = this.#cliArgs;
-
-		if (!group || group === 'default') {
-			if (playwright) {
-				console.warn('Warning: reducedMotion disabled. Use the unit group to enable reducedMotion.');
-			} else {
-				console.warn('Warning: Running with puppeteer, reducedMotion disabled. Use the unit group to use playwright with reducedMotion enabled');
-			}
-		}
```
bearfriend, line 166: No longer necessary since we are disabled `default` and group and defaulting to `unit` (and forcing all `browsers` options to playwright browsers).

bin/test.js
```diff
@@ -179,11 +181,10 @@ export class WTRConfig {
 		config.groups ??= [];
 		config.groups.push({
 			name: 'unit',
-			files: this.#getPattern('test'),
-			browsers: this.getBrowsers()
+			files: this.#getPattern('test')
 		});
 
-		if (vdiff) {
+		if (group === 'vdiff') {
```
bearfriend, line 187: No more `vdiff: true`. Just use the group to enable it.

src/server/index.js
bearfriend, line 1: delete me

src/server/visual-diff-plugin.js
```diff
@@ -89,6 +89,7 @@ async function tryMoveFile(srcFileName, destFileName) {
 		await rename(srcFileName, destFileName);
 		return true;
 	} catch (e) {
+		console.log('HEY!');
```
bearfriend, line 92: lol 🧹

src/server/visual-diff-reporter.js
```diff
@@ -112,7 +112,7 @@ export function visualDiffReporter({ reportResults = true } = {}) {
 			cpSync(inputDir, tempDir, { force: true, recursive: true });
 			writeFileSync(join(tempDir, 'data.js'), `export default ${json};`);
 
-			execSync(`rollup -c ${join(__dirname, './rollup.config.js')}`);
+			execSync(`npx rollup -c ${join(__dirname, './rollup.config.js')}`, { stdio: 'pipe' });
```
bearfriend, line 115: `npx` fixes `command not found` error, which I believe is necessary now the way this runs.

`stdio` fixes the double output, though I'm not sure why, since it's supposed to be the default. May need to make this `ignore` if it spits out other bad things.

bearfriend:
Tests in progress

bin/test.js
dlockhart, line 1: Thoughts on this being `d2l-test-runner` to make its purpose a bit more explicit and leave the space open for other `d2l-test-*` things in the future? The config file would then become `d2l-test-runner.config.js`.

bin/test.js
```diff
@@ -0,0 +1,151 @@
...
+
+const DISALLOWED_OPTIONS = ['--browsers', '--playwright', '--puppeteer', '--groups'];
+
+const optionDefinitions = [
+	// @web/test-runner options
+	{
+		name: 'files',
+		type: String,
+		multiple: true,
+		description: 'Test files to run. Path or glob.\n[Default: ./test/**/*.<type>.js]',
```
dlockhart, line 17:
```suggestion
		description: 'Test files to run. Path or glob.\n[Default: ./test/**/*.<group>.js]',
```

src/server/visual-diff-reporter.js
```diff
@@ -1,8 +1,8 @@
-import { cpSync, mkdirSync, rmSync, writeFileSync } from 'fs';
-import { dirname, join } from 'path';
+import { cpSync, mkdirSync, rmSync, writeFileSync } from 'node:fs';
+import { dirname, join } from 'node:path';
```
dlockhart, line 2: I'm always unsure when to use `fs` vs. `node:fs`... seems like you know more about this!

bin/test.js
bearfriend: We could do that, or we could make this `d2l-test run` as a subcommand and future things could have their own subcommands. Even `migrate` could live here as `d2l-test migrate`.

svanherk: We'd thrown around wrapping the dev server at one point as well.  Would you imagine that being a totally separate command like `d2l-dev-server` or whatever?  Or also a subcommand? Because then we might want something more generic like `d2l-ci serve`, `d2l-ci test`, etc.  Migrate is the same question (`d2l-test migrate` or `d2l-vdiff-migrate`), so just trying to figure out the pattern now

bin/test.js
svanherk: Neither of these are _really_ accurate, right? For group `unit`, this is `test`.  For group `vdiff`, this is `vdiff`.

bin/test.js
```diff
@@ -0,0 +1,151 @@
...
+	{
+		name: 'group',
+		type: String,
+		required: true,
+		defaultOption: true,
+		description: 'Name of the group to run tests for\n[Default: unit]',
+		order: 1
+	},
+	{
+		name: 'manual',
```
svanherk, line 29: As discussed in standup, maybe we remove this option to discourage it, and people like me that still like it can just go `--watch`, then `m`, then `d` (which I already do now)

bin/test.js
```diff
@@ -0,0 +1,151 @@
...
+	{
+		name: 'firefox',
+		type: Boolean,
+		description: 'Run tests in Firefox',
+		order: 3
+	},
+	{
+		name: 'golden',
+		type: Boolean,
+		description: 'Generate new golden screenshots',
```
svanherk, line 72:
```suggestion
		description: 'Generate new golden screenshots when using group vdiff',
```
Or whatever

test/browser/d2l-test.config.js
```diff
@@ -13,8 +10,7 @@ function getGoldenFlag() {
 	};
 }
 
-export default createConfig({
-	pattern,
-	vdiff: true,
+export default {
+	pattern: type => `test/browser/**/*.${type}.js`,
```
svanherk, line 14:
```suggestion
	pattern: () => `test/browser/**/*.vdiff.js`,
```
Technically this could just be this, since we only use this config file with our vdiff tests here?  Can they name a config file whatever they want and we make that more explicit (`vdiff-config.js`)?

src/server/visual-diff-reporter.js
bearfriend: I think they're supposed to do the same thing, but the `node:` prefix is newer and makes clear where it's coming from, instead of looking like a package. However, I added these because I was getting an error at some point, but didn't dig deeper since this fixed it.

bin/test.js
bearfriend: Closing the loop. We decided the `unit` group was weird anyway, and it should just be a generic `test` group, which makes using `<group>` here accurate.

test/browser/d2l-test.config.js
```diff
@@ -13,8 +10,7 @@ function getGoldenFlag() {
 	};
 }
 
-export default createConfig({
-	pattern,
-	vdiff: true,
+export default {
+	pattern: () => `test/browser/**/*.vdiff.js`,
```
bearfriend, line 14:
```suggestion
	pattern: () => 'test/browser/**/*.vdiff.js',
```

package.json
```diff
@@ -7,12 +7,16 @@
...
     "start": "web-dev-server --root-dir ./.vdiff --open ./.report/",
-    "test": "npm run lint && npm run test:server && npm run test:browser",
-    "test:browser": "web-test-runner --files \"./test/browser/**/*.test.js\" --node-resolve --playwright",
-    "test:server": "mocha ./test/server/**/*.test.js",
-    "test:vdiff": "web-test-runner --config ./test/browser/wtr-vdiff.config.js --group vdiff",
+    "test": "npm run lint && npm run test:bin && npm run test:server && npm run test:browser",
+    "test:bin": "mocha './test/bin/**/*.test.js'",
+    "test:browser": "web-test-runner --files './test/browser/**/*.test.js' --node-resolve --playwright",
+    "test:server": "mocha './test/server/**/*.test.js'",
+    "test:vdiff": "d2l-test --config ./test/browser/d2l-test.config.js --group vdiff",
```
dlockhart, line 14: I still get a `d2l-test: command not found` when I run `npm run test:vdiff`. If I change this line to `npx` it works.

test/server/cli/test.test.js
```diff
@@ -0,0 +1,56 @@
...
+	});
+
+	it('should use --group for the default option', async() => {
+		const opts = await runner.getOptions(['group-name']);
+		expect(opts.argv).to.deep.equal(['--group', 'group-name']);
+	});
+
+	it('should not forward disallowed or stolen options', async() => {
+		const disallowed = ['--browsers', '--playwright', '--puppeteer', '--groups'];
+		const stolen = ['--config'];
```
dlockhart, line 28: 😆 stolen

src/server/cli/test.js
```diff
@@ -0,0 +1,159 @@
...
+		{
+			name: 'group',
+			type: String,
+			required: true,
+			defaultOption: true,
+			description: 'Name of the group to run tests for\n[Default: test]',
+			order: 1
+		},
+		{
+			name: 'manual',
```
svanherk, line 30: Not sure if I raised this here or the other PR, this will still be removed in a future PR?

src/server/cli/test.js
```diff
@@ -0,0 +1,159 @@
...
+		{
+			name: 'watch',
+			type: Boolean,
+			description: 'Reload tests on file changes. Allows debugging in all browsers.',
+			order: 9
+		},
+
+		// d2l-test options
+		{
+			name: 'chrome',
```
svanherk, line 44: One thing I was thinking about - does this affect the browser names the runner has?  Will the goldens be stored in `chrome` and `safari` folders, or still `chromium` and `webkit`?

src/server/wtr-config.js
```diff
@@ -125,44 +131,17 @@ export class WTRConfig {
...
-			if (playwright) {
-				console.warn('Warning: reducedMotion disabled. Use the unit group to enable reducedMotion.');
-			} else {
-				console.warn('Warning: Running with puppeteer, reducedMotion disabled. Use the unit group to use playwright with reducedMotion enabled');
-			}
+		if (!['test', 'vdiff', ...passthroughGroupNames].includes(group)) {
+			return {}; // allow wtr to error
 		}
 
 		delete passthroughConfig.browsers;
```
svanherk, line 147: Do we still need this if we never send this along? Or is this in case they set `browsers` in their config file?

test/server/cli/test.test.js
bearfriend: Hah. I tried to avoid it at first but then when I got to `borrowed` it seemed like the most accurate word.

src/server/wtr-config.js
bearfriend: Yeah, the latter.

src/server/cli/test.js
bearfriend: Nothing else changes. I did start second-guessing this because literally everything else will not match these. Directories, test output, report tabs, (playwright/wtr options).

src/server/cli/test.js
```diff
@@ -0,0 +1,159 @@
...
+			multiple: true,
+			description: 'Test files to run. Path or glob.\n[Default: ./test/**/*.<group>.js]',
+			order: 8
+		},
+		{
+			name: 'group',
+			type: String,
+			required: true,
+			defaultOption: true,
+			description: 'Name of the group to run tests for\n[Default: test]',
```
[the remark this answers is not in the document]
bearfriend: I used the D2L rule of one sentence == no period, multiple sentences == periods, but I did miss a couple. Are we ok with that rule or do we always want periods?

package.json
bearfriend: Yeah it's been working for me because I have it linked which creates a global symlink, but we can add `npx` here. I kind of don't like it (maybe just aesthetically) but I didn't find any more official solution to this problem with a quick search so :shrug:

bin/test.js
bearfriend: I did mean to come back to this. Right now I'm thinking `d2l-test` or `d2l-test run` do the same thing, and then we can add `d2l-test migrate-vdiff`, and others. Maybe `d2l-test report` to start a server and launch the report? Let's chat offline.

src/server/cli/test.js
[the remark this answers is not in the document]
dlockhart: I was unaware of this rule!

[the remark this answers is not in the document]
svanherk: Was also unaware of this rule! I'm fine with that, but yeah I couldn't find a pattern in what was currently there

src/server/cli/test.js
svanherk: Yeah, it's convenient but makes me like it a bit less if everything else will still say the old name

dlockhart: I'd be happy to switch what the report shows if that's what we're worried about. The output in the console though will still say "chromium" and "webkit". Surprised Firefox doesn't resort to "gecko".

svanherk: I think the weirdest part for me would be the folders not matching, which I guess we could change by having a map in the plugin.

test/browser/d2l-test.config.js
svanherk, line 1: I still like the idea of this file being called `config.vdiff.js` or something lol, just to imply it's not used for the other tests in this folder

src/server/cli/test.js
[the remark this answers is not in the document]
bearfriend: [For reference](https://daylight.d2l.dev/style-elements/writing/#periods)

package.json
```diff
@@ -7,12 +7,16 @@
...
     "start": "web-dev-server --root-dir ./.vdiff --open ./.report/",
-    "test": "npm run lint && npm run test:server && npm run test:browser",
-    "test:browser": "web-test-runner --files \"./test/browser/**/*.test.js\" --node-resolve --playwright",
-    "test:server": "mocha ./test/server/**/*.test.js",
-    "test:vdiff": "web-test-runner --config ./test/browser/wtr-vdiff.config.js --group vdiff",
+    "test": "npm run lint && npm run test:bin && npm run test:server && npm run test:browser",
+    "test:bin": "mocha './test/bin/**/*.test.js'",
+    "test:browser": "web-test-runner --files './test/browser/**/*.test.js' --node-resolve --playwright",
+    "test:server": "mocha './test/server/**/*.test.js'",
+    "test:vdiff": "npx d2l-test-runner--config ./test/browser/d2l-test.config.js --group vdiff",
```
svanherk, line 14:
```suggestion
    "test:vdiff": "npx d2l-test-runner--config ./test/browser/config.vdiff.js --group vdiff",
```

dlockhart (approved):

svanherk (approved):

PR #66, opened 2023-07-11
render browsers as tabs in the the header

dlockhart:
Latest updates renders a tab for each browser in the right side's header, and introduces a floating footer for additional information about the browser and the selected test result. The back button is finally in the top-right, where I'll add next/previous iterators shortly.

Screenshot 2023-07-11 at 3 40 05 PM


src/server/report/app.js
```diff
@@ -217,15 +229,18 @@ class App extends LitElement {
 			`;
 		};
 
+		const browserFilter = data.browsers.length > 1 ? html`
+			<fieldset>
+				<legend>Browsers</legend>
+				${ data.browsers.map(b => renderBrowser(b))}
+			</fieldset>` : nothing;
```
dlockhart, line 236: Realized I was a bit eager with testing all 3 browsers and hadn't really been thinking about the more common case of just one. This hides the browser filter entirely if there's only 1 to filter on.

src/server/report/test.js
```diff
@@ -142,6 +295,77 @@ class Test extends LitElement {
...
+					aria-controls="tabpanel-${browser}"
+					aria-selected="${selected ? 'true' : 'false'}"
+					id="tab-${browser}"
+					role="tab"
+					tabindex="${selected ? '0' : '-1'}"
+					type="button"
+					@click="${onClick()}">
+						<span>${browser} <span class="${classMap(statusClass)}">(${status})</span></span>
+						${selected ? html`<div class="tab-selected-indicator"></div>` : nothing}
+				</button>`;
```
bearfriend, line 342: More of a question than a suggestion. I got into the habit of formatting like this because I found it easier to follow the markup, and easy _enough_ to then find and follow the javascript within/around it, which overall made it easier to find and read than, for example, trying to scan for that `<div>` in there.

Admittedly this looks pretty terrible in the github comment UI, though, and maybe it's just a messy example anyway.
```suggestion
				<button
					aria-controls="tabpanel-${browser}"
					aria-selected="${selected ? 'true' : 'false'}"
					id="tab-${browser}"
					role="tab"
					tabindex="${selected ? '0' : '-1'}"
					type="button"
					@click="${onClick()}">
						<span>
							${browser} <span class="${classMap(statusClass)}">(${status})</span>
						</span>${selected ? html`
						<div class="tab-selected-indicator"></div>` : nothing}
				</button>`;
```

src/server/report/button.js
```diff
@@ -0,0 +1,40 @@
...
+			display: flex;
+			gap: 5px;
+			line-height: 24px;
+			margin: 0;
+			outline: none;
+			padding: 10px;
+			user-select: none;
+		}
+		button:hover,
+		button:focus-visible {
```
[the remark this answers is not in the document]
dlockhart: Yeah you're right that works here since the hover/focus styles are the same (for now) -- updated.

src/server/report/test.js
dlockhart: Sure why not!

src/server/report/test.js
```diff
@@ -121,15 +236,54 @@ class Test extends LitElement {
...
+		}
+		if (selectedBrowserIndex < 0) {
+			selectedBrowserIndex = browsers.findIndex(b => testData.results.find(r => r.name === b.name && !r.passed));
+		}
+		if (selectedBrowserIndex < 0) {
+			selectedBrowserIndex = 0;
+		}
+		const selectedBrowser = browsers[selectedBrowserIndex];
+		return selectedBrowser;
+	}
```
bearfriend, line 258:
```suggestion
	_getSelectedBrowser(browsers, { results }) {
		return browsers[this._selectedBrowserIndex] ||
			browsers.find(b => results.find(r => r.name === b.name && !r.passed)) ||
			browsers[0];
	}
```

dlockhart: Forgot that you can access out of bounds array elements and it doesn't throw. This is... much less verbose!

bearfriend (approved):

PR #68, opened 2023-07-12
improvements to the diff panel

dlockhart:
Some mostly minor improvements to the area where the actual diff is shown:
- Blue dotted divider between the split view
- Shows the name (golden vs. new)
- Shows the width and height

Screenshot 2023-07-12 at 3 06 45 PM

I decided that having a separate web component for the test result wasn't working out, so I've merged it into the test one. As a follow-up, I think I'm going to merge the test one into the app as well and split styles and render helpers out into separate files. Ultimately here there was no advantage to having separate components if they're only used from one place -- lesson learned.


svanherk (approved):

src/server/report/test.js
```diff
@@ -185,6 +184,71 @@ class Test extends LitElement {
...
+		}
+		.result-part-info {
+			align-items: center;
+			display: flex;
+			gap: 5px;
+		}
+		.result-part-info-spacer,
+		.result-part-info-size {
+			flex: 1 0 0%;
+			text-align: right;
```
svanherk, line 236: I think I like this slightly better `text-align: left;` (and `gap` reduced) but I see the benefit of having both at the ends

svanherk:
Oh, also just noticed the passed tests are looking a little squished lol: 
image
I think I'd expect the `golden` to have the image, and `new` to have "No Changes"?

dlockhart:
Oooh that is squishy and ugly! Yeah maybe it makes sense to flip those around... I'll fix this stuff up thanks!

PR #69, opened 2023-07-12
refactor tabs and radio stuff out

dlockhart:
No visual or functional changes here.

This moves all the tabs and pillbox related rendering and CSS out into separate helper files. This will help reduce the size of `test.js` in preparation for merging it back into `app.js` in part 2 of the refactoring. Because Rollup builds all this together, there's no impact on the report output files -- still just `index.html` and `app.js`.

Background: I originally split these into separate components but in since they don't really need to be reused they're creating overhead to pass stuff down and communicate up.

### DIFF
--- a/src/server/report/radio.js
+++ b/src/server/report/radio.js
@@ -1,0 +1,63 @@
+import { css, html } from 'lit';
+import { getId } from './common.js';
+
+export const RADIO_STYLE = css`
+	.radio-container {
+		border-radius: 5px;
+		display: flex;
+		flex-wrap: nowrap;
+	}
+	.radio-item input[type="radio"] {
+		position: absolute;
+		opacity: 0;
+		pointer-events: none;
+	}
+	.radio-item label {
+		align-items: center;
+		background-color: #ffffff;
+		border-block-end-style: solid;
+		border-block-start-style: solid;
+		border-inline-start-style: solid;
+		border-color: #cdd5dc;
+		border-width: 1px;
+		cursor: pointer;
+		display: flex;
+		gap: 5px;
+		line-height: 24px;
+		padding: 10px;
+		position: relative;
+		user-select: none;
+	}
+	.radio-item:first-child label {
+		border-start-start-radius: 5px;
+		border-end-start-radius: 5px;
+	}
+	.radio-item:last-child label {
+		border-end-end-radius: 5px;
+		border-inline-end-style: solid;
+		border-start-end-radius: 5px;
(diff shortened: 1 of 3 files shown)


src/server/report/test.js
```diff
@@ -339,83 +268,20 @@ class Test extends LitElement {
...
-					${this._renderTestResults(browser, testData)}
-				</div>
-			`;
-		};
+		let panelsContent;
+		if (tabs.length < 2) {
+			panelsContent = tabs.map(t => t.content);
+		} else {
+			panelsContent = tabs.map(t => renderTabPanel(t));
+		}
```
dlockhart, line 282: Slight bug fix here -- if there's only 1 browser showing we don't render the tabs at all, so this also skips rendering the surrounding panel.

src/server/report/radio.js
```diff
@@ -0,0 +1,63 @@
+import { css, html } from 'lit';
+import { getId } from './common.js';
+
+export const RADIO_STYLE = css`
```
svanherk, line 4: > No visual or functional changes here.

Are the radio styles new?  Or am I losing it, where did they come from before?

svanherk: Nvm, I am losing it.  They were previously "pills" lol

dlockhart: Haha yeah I took the opportunity to rename them to radio.

src/server/report/test.js
```diff
@@ -464,8 +330,7 @@ class Test extends LitElement {
 		}
 
 	}
-	_renderTestResults(browser, testData) {
-		const result = testData.results.find(r => r.name === browser);
+	_renderTestResults(result) {
```
svanherk, line 333: Seems like this is maybe overkill now, and the `div` could just be added to `_renderTestResult`?

svanherk: I guess there's a lot of early `return`s in there.  Could maybe move this up into:
```
content: html`<div class="result">this._renderTestResults(result)</div>`,
```
Either way, minor, just an idea

svanherk (approved):
I like this pattern of still logically separating out pieces, but without the overhead that comes with components and data passing if we don't need it

src/server/report/test.js
dlockhart: Yeah I had started leaving space for when this is actually going to be rendering multiple test results -- i.e. once you can click on an entire file and see all its tests. Definitely overkill right now though.

PR #70, opened 2023-07-12
fix case where golden and new match

dlockhart:
Some follow-up fixes to #68 based on feedback from Stacey.

Sizes are now left aligned but the names are still centred:
Screenshot 2023-07-12 at 5 12 43 PM

Fixes the case where the goldens match the new and it was super squished and ugly. Also flipped it around so the "no changes" message shows on the right side.
Screenshot 2023-07-12 at 5 11 12 PM


svanherk (approved):
Haha love it! Now I want a collection of happy images that are randomly selected 😝

PR #71, opened 2023-07-13
Add fixture pausing and --open option to replace --manual

bearfriend:
[US152298](https://rally1.rallydev.com/#/?detail=/userstory/699970327679&fdp=true): Visual-diff: ability to debug in the browser

- Removes `--manual` option (still accessible via `--watch`)
- Adds `--open` option to run tests in headed playwright browsers without `--watch`
- Fixes various issues with taking screenshots in headed Chrome and both headed and headless Firefox
- Adds `fixture` pausing in headed browser modes with UI controls to
  - start or skip test files
  - run or skip individual tests
  - retry failed tests
  - run all remaining tests


package.json
```diff
@@ -33,7 +33,8 @@
     "sinon": "^15"
   },
   "exports": {
-    ".": "./src/browser/index.js"
+    ".": "./src/browser/index.js",
+    "./pause.js": "./src/server/pause.js"
```
dlockhart, line 37: It's unfortunate that we need to export this even though consumers don't need to know it exists. Wondering if we could just inject it into the HTML document that `wtr-config.js` generates, similar to the ResizeObserver error handler and visual-diff styles.

src/server/pause.js
```diff
@@ -0,0 +1,42 @@
...
+			<span>.${window.__WTR_CONFIG__.testFile.split('?')[0]}</span>
+		</div>
+		<div id="run" hidden>
+			<button id="run-button" style="margin-inline-end: 1rem;">Run</button>
+			<span id="test-name"></span>
+		</div>
+	</div>
+`;
+
+document.documentElement.insertAdjacentHTML('afterBegin', controls);
```
dlockhart, line 16: When I inspect this, it's actually getting injected _above_ the `<head>` and not even inside the `<body>`. That seems... wrong.

As an aside, I'm wondering if we should make it fixed position at the bottom of the screen so that it's less likely to overlap or interfere with things under test?

src/server/pause.js
```diff
@@ -0,0 +1,42 @@
...
+});
+
+function run() {
+	runBtn.disabled = true;
+	window.d2lTestRun();
+}
+
+function start() {
+	document.querySelector('#start').hidden = true;
+	document.querySelector('#run').hidden = false;
```
dlockhart, line 37: The focus management here is a bit weird, but maybe there isn't much we can do. I'd expect to want to hit "start", move focus to "run" and then keep it there and hit ENTER a bunch of times until it gets to the test I want to see.

src/server/wtr-config.js
```diff
@@ -193,9 +194,14 @@ export class WTRConfig {
...
 		return browsers.map((b) => playwrightLauncher({
-			concurrency: b === 'firefox' ? 1 : undefined, // focus in Firefox unreliable if concurrency > 1 (https://github.com/modernweb-dev/web/issues/238)
+			concurrency: b === 'firefox' || this.#cliArgs.open ? 1 : undefined, // focus in Firefox unreliable if concurrency > 1 (https://github.com/modernweb-dev/web/issues/238)
 			product: BROWSER_MAP[b],
-			createBrowserContext: ({ browser }) => browser.newContext({ deviceScaleFactor: 2, reducedMotion: 'reduce' })
+			createBrowserContext: ({ browser }) => browser.newContext({ deviceScaleFactor: 2, reducedMotion: 'reduce' }),
+			launchOptions: {
+				headless: !this.#cliArgs.open,
+				devtools: false,
+				slowMo: this.#cliArgs.slowmo || 0
```
dlockhart, line 203: Maybe I'm missing it, but I don't see any code that actually does anything when `slowMo` is set?

src/server/headed-mode-plugin.js
```diff
@@ -1,16 +1,17 @@
...
 	return {
 		name: 'brightspace-headed-mode',
 		async transform(context) {
-			if ((watch || manual) && files.includes(context.path.slice(1))) {
-				// allow time to open devtools in firefox and webkit
-				watch && await new Promise(r => setTimeout(r, 2000));
-				return `debugger;\n${context.body}`;
+			if ((watch || open) && files.includes(context.path.slice(1))) {
+				await delay(0);
+				return `debugger;\nimport '@brightspace-ui/testing/pause.js';\n${context.body}`;
```
dlockhart, line 14: Do we still need to pause the debugger now that there's a "start" button?

src/server/pause.js
bearfriend: I did have it fixed to the bottom, but I realized that could actually interfere more, I think. Like if the screenshot is finding the fixture rectangle, it will always be below this, but fixed, it could overlap, right?

src/server/pause.js
bearfriend: Yeah the tests will move focus all over,  but we can always move it back to `run` before we pause.

src/server/wtr-config.js
bearfriend: It's built into playwright so it slows down operations, though I'm sure what all because they still go pretty fast.

dlockhart: Aha makes sense!

src/server/headed-mode-plugin.js
bearfriend: I think so. Pausing gives you time to open the devtools, and then starting will plop you right on the test file because of `debugger;` where you can then set other breakpoints. Also useful because the browser has no persistence and you lose all your devtools state.

src/server/pause.js
dlockhart: Right. What just inserting it as the first child of the body work? Just don't like seeing it above the `<head>`.

src/browser/vdiff.js
```diff
@@ -1,28 +1,106 @@
...
 	rootHooks: {
 		beforeEach() {
 			test = this.currentTest;
 		}
 	}
 });
+/* eslint-enable */
 
 async function ScreenshotAndCompare(opts) {
+	await document.fonts.ready; // firefox fonts
```
bearfriend, line 44: For some reason Firefox fonts go back into `loading`, so we need to double check they are done before the screenshot.

src/browser/vdiff.js
```diff
@@ -1,28 +1,106 @@
...
+
+	window.d2lTest?.pass();
+}
+
+const disallowedProps = ['width', 'inline-size'];
+let count = 0;
+function inlineStyles(elem) {
+	// headed chrome takes screenshots by first moving the element, which
+	// breaks the hover state. So, copy current styles inline before screenshot
+	if (window.d2lTest.hovering && window.chrome) {
```
dlockhart, line 77: Nit: invert this and return to reduce nesting

src/browser/vdiff.js
```diff
@@ -1,28 +1,106 @@
...
 	}
+
+	window.d2lTest?.pass();
+}
+
+const disallowedProps = ['width', 'inline-size'];
+let count = 0;
+function inlineStyles(elem) {
+	// headed chrome takes screenshots by first moving the element, which
+	// breaks the hover state. So, copy current styles inline before screenshot
```
dlockhart, line 76: Whoa, crazy that we'd need to do this. I'm sure you tried this, but instead of making all the styles inline, could we just try to re-apply the hover after Chrome moves it?

bearfriend: Yeah, it's not _just_ that it moves. It seems to be locking it down in some way, which makes sense, but it just loses the hover.

src/browser/vdiff.js
```diff
@@ -1,28 +1,106 @@
...
-chai.Assertion.addMethod('golden', function(...args) {
-	return ScreenshotAndCompare.call({ test, elem: this._obj }, ...args); // eslint-disable-line no-invalid-this
+/* eslint-disable no-undef, no-invalid-this */
+chai.Assertion.addMethod('golden', async function(...args) {
+	await soonPromise?.catch(err => expect.fail(err));
+	return ScreenshotAndCompare.call({ test, elem: this._obj }, ...args);
 });
-mocha.setup({ // eslint-disable-line no-undef
+
+chai.Assertion.addChainableMethod('soon',
```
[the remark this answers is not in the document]
bearfriend: Yeah, that's totally fair. The (mostly) common thread was changing things outside the component itself that would cause it to update (like scrolling or inserting HTML into the fixture's light DOM), but not until the next frame, so `updateComplete` isn't necessarily doing anything helpful here, but seemed good to include, except when you consider the potential for misuse.

Anyway, I think we can get rid of it and change these `floating-buttons` test like this:
```javascript
	// fixed with oneEvent()
	it('does not float at bottom of container', async() => {
		const elem = await fixture(floatingButtonsFixture);
		window.scrollTo(0, document.body.scrollHeight);
		await oneEvent(window, 'scroll');
		await expect(elem).to.be.golden();
	});

	// doesn't make sense to me that this one would need a nextFrame(), but it does fix it
	it('does not float when small amount of content', async() => {
		const elem = await fixture(floatingButtonsShortFixture);
		await nextFrame();
		await expect(elem).to.be.golden();
	});

	// fixed with nextFrame()
	it('floats when content added to dom', async() => {
		const elem = await fixture(floatingButtonsShortFixture);
		const contentElem = document.querySelector('#floating-buttons-short-content').querySelector('p');
		contentElem.innerHTML += '<br><br>I love Coffee...';
		await nextFrame();
		await expect(elem).to.be.golden();
	});
```

To be honest I kind of just wanted to get a custom chai property to work 😆

src/browser/vdiff.js
bearfriend: Yeah it should be.

src/browser/vdiff.js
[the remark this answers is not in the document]
dlockhart: Haha cool. Yeah I'm totally fine with tests needing to `await nextFrame()` themselves if they do things that would require waiting for a repaint. At least that's logical.

bearfriend:
Since the base (#65) got reworked the conflicts are a bit much, and this is a doing lots of things so I'll just close this and re-open the individual bits as their own PRs.

PR #78, opened 2023-07-18
ability to render multiple test results

dlockhart:
This re-arranges everything so that you can now click on an entire file and see all the tests within that file, split out by browser. Looks like this:

Screenshot 2023-07-18 at 5 22 02 PM

I removed the footer since it was taking up a lot of space and am just rendering the browser version info at the top of the test list.


src/server/report/app.js
```diff
@@ -251,32 +207,22 @@ class App extends LitElement {
...
 	}
-	_renderFile(file) {
-		const renderBrowserCell = (b) => {
-			if (!this._filterBrowsers.includes(b.name)) return nothing;
-			return html`<th>${b.name}</th>`;
-		};
+	_handleOverlayChange(e) {
+		this._overlay = e.target.checked;
+	}
+	_renderEmpty() {
```
dlockhart, line 217: Hopefully Jeff can improve on this... 

Screenshot 2023-07-18 at 5 22 15 PM

bearfriend: Not possible

src/server/report/app.js
```diff
@@ -369,12 +303,55 @@ class App extends LitElement {
...
 			</tr>
 		`;
 	}
-	_renderTestResults(fileData, testData) {
+	_renderMainView() {
+
+		if (this._filterFile === undefined) {
+			let list;
+			if (this._files.length === 0) {
+				list = html`<p>No tests exist for the selected filters.</p>`;
```
bearfriend, line 316: Does this work here? I want more sad face magnifying glass.
```suggestion
				return this._renderEmpty();
```
If so, then probably some changes to the rest of this block.

dlockhart: Thanks good catch -- yep more sad magnifying glass was required.

src/server/report/app.js
```diff
@@ -387,20 +364,29 @@ class App extends LitElement {
...
+			const numPassed = browserResults.get(b.name);
 			return {
-				content: html`<div style="padding: 20px;">${renderResult(result, { fullMode: this._fullMode, layout: this._layout, showOverlay: this._overlay })}</div>`,
+				content: renderBrowserResults(b, tests, { fullMode: this._fullMode, layout: this._layout, showOverlay: this._overlay }),
 				label: b.name,
 				id: b.name.toLowerCase(),
 				selected: b.name === selectedBrowser.name,
-				status: result.passed ? 'passed' : 'failed',
-				statusType: result.passed ? TAB_STATUS_TYPE.NORMAL : TAB_STATUS_TYPE.ERROR
+				status: `${numPassed}/${tests.length}`,
```
svanherk, line 388:
```suggestion
				status: `${numPassed}/${tests.length} passed`,
```
I wonder if this would be helpful, or if it's just something you need to get used to.  But when this is red, I couldn't immediately tell if it was telling me number of failed tests or number of passed tests

dlockhart: I was wondering about that too -- let's add it for extra clarity and we can always adjust later.

Screenshot 2023-07-19 at 10 23 50 AM

src/server/report/result.js
```diff
@@ -121,3 +160,35 @@ export function renderResult(resultData, options) {
...
+					<h3>${t.name}</h3>
+					<div class="${classMap(durationClass)}">${duration}ms</div>
+				</div>
+				${renderResult(resultData, options)}
+			</div>
+		`;
+	});
+	return html`
+		<div class="result-browser padding">
+			${ICON_BROWSERS[browser.name]}
```
svanherk, line 186: To save more space, is there a reason the icon and version don't just live in the tab?  I'm guessing because of the case where there's only one browser and are no tabs?

svanherk: Yeah if there's more coming to the area that makes sense, would be interested in Jeff's thoughts/ideas on it eventually

svanherk (approved):

PR #80, opened 2023-07-19
Filter test results to match status

dlockhart:
A few tweaks here based on feedback:

- When viewing an entire file's tests, browser results for those tests should be filtered according to the pass/fail/all status
- When switching between tabs, scroll to the top
- Centralized the rendering of `(x/y passed)`

### DIFF
--- a/src/server/report/app.js
+++ b/src/server/report/app.js
@@ -1,11 +1,11 @@
 import './button.js';
+import { COMMON_STYLE, FILTER_STATUS, FULL_MODE, LAYOUTS, renderEmpty, renderTestStatus } from './common.js';
 import { css, html, LitElement, nothing } from 'lit';
-import { FILTER_STATUS, FULL_MODE, LAYOUTS } from './common.js';
-import { ICON_EMPTY, ICON_HOME } from './icons.js';
 import { RADIO_STYLE, renderRadio } from './radio.js';
 import { renderBrowserResults, RESULT_STYLE } from './result.js';
-import { renderTabButtons, renderTabPanel, TAB_STATUS_TYPE, TAB_STYLE } from './tabs.js';
+import { renderTabButtons, renderTabPanel, TAB_STYLE } from './tabs.js';
 import data from './data.js';
+import { ICON_HOME } from './icons.js';
 import page from 'page';
 
 class App extends LitElement {
@@ -18,7 +18,7 @@ class App extends LitElement {
 		_overlay: { state: true },
 		_selectedBrowserIndex: { state: true }
 	};
-	static styles = [RADIO_STYLE, RESULT_STYLE, TAB_STYLE, css`
+	static styles = [COMMON_STYLE, RADIO_STYLE, RESULT_STYLE, TAB_STYLE, css`
 		.container {
 			display: grid;
 			grid-auto-flow: row;
@@ -73,26 +73,6 @@ class App extends LitElement {
 		}
 		fieldset > legend {
 			font-weight: bold;
-		}
-		.empty {
-			align-items: center;
-			display: flex;
-			flex-direction: column;
-			gap: 20px;
-		}
-		.empty > svg {
-			color: #007bff;
-			height: 100px;
-			width: 100px;
-		}
-		.empty > p {
-			color: #6e7477;
-			font-size: 1.1rem;
-			font-weight: bold;
-			margin: 0;
-		}
-		.padding {
-			padding: 20px;
 		}
 		.test-results {
 			display: grid;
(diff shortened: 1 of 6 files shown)


src/server/report/app.js
```diff
@@ -243,11 +216,10 @@ class App extends LitElement {
 
 		const renderBrowser = (b) => {
 			const browserData = data.browsers.find(data => data.name === b.name);
-			const result = `${(data.numTests - browserData.numFailed)}/${data.numTests}`;
 			return html`
 				<label>
 					<input type="checkbox" value="${b.name}" ?checked="${this._filterBrowsers.includes(b.name)}" @click="${this._handleFilterBrowserChange}">
-					${b.name} (${result} passed)
+					${b.name} ${renderTestStatus(data.numTests - browserData.numFailed, data.numTests)}
```
dlockhart, line 222: Screenshot 2023-07-19 at 4 07 12 PM

src/server/report/result.js
```diff
@@ -189,6 +187,6 @@ export function renderBrowserResults(browser, tests, options) {
 				<div>version ${browser.version}</div>
 			</div>
 		</div>
-		${results}
+		${filteredTests.length === 0 ? renderEmpty() : results}
```
dlockhart, line 190: Since a browser's results can now be empty, one more place for sad magnifying glass!

Screenshot 2023-07-19 at 4 02 00 PM

src/server/report/result.js
```diff
@@ -162,20 +153,27 @@ function renderResult(resultData, options) {
...
-			'pass': duration < 500,
-			'warning': duration >= 500 && duration < 1000
-		};
+		if (resultData.passed && options.filterStatus === FILTER_STATUS.FAILED ||
+			!resultData.passed && options.filterStatus === FILTER_STATUS.PASSED) {
+			return false;
+		}
+		return true;
+	});
+	const results = filteredTests.map(t => {
```
bearfriend, line 164: Instead of looping through `tests` and `results` twice each, I'd use a `reduce`
```javascript
export function renderBrowserResults(browser, tests, options) {
	const results = tests.reduce((acc, t) => {
		const resultData = t.results.find(r => r.name === browser.name);
		if (resultData.passed && options.filterStatus === FILTER_STATUS.FAILED ||
			!resultData.passed && options.filterStatus === FILTER_STATUS.PASSED) {
			return acc;
		}

		let status = STATUS_TYPE.NORMAL;
		if (resultData.duration >= 1000) {
			status = STATUS_TYPE.ERROR;
		} else if (resultData.duration >= 500) {
			status = STATUS_TYPE.WARNING;
		}

		return acc.push(html`
			<div class="result-container">
				<div class="result-test-name">
					<h3>${t.name}</h3>
					<div class="result-duration">${renderStatusText(`${resultData.duration}ms`, status)}</div>
				</div>
				${renderResult(resultData, options)}
			</div>
		`) && acc;
	}, []);
	return html`
		<div class="result-browser padding">
			${ICON_BROWSERS[browser.name]}
			<div>
				<div class="result-browser-name">${browser.name}</div>
				<div>version ${browser.version}</div>
			</div>
		</div>
		${results.length === 0 ? renderEmpty() : results}
	`;
}
```

dlockhart: That's a really elegant way to filter & map at the same time -- thanks!

src/server/report/result.js
```diff
@@ -162,25 +153,32 @@ function renderResult(resultData, options) {
...
-			'warning': duration >= 500 && duration < 1000
-		};
-		return html`
+		if (resultData.passed && options.filterStatus === FILTER_STATUS.FAILED ||
+			!resultData.passed && options.filterStatus === FILTER_STATUS.PASSED) {
+			return acc;
+		}
+
+		let status = STATUS_TYPE.NORMAL;
+		if (resultData.duration >= 1000) {
```
bearfriend, line 165: We can do this later if we want to at all, but can we hook this value up to the `slow` mocha config?

bearfriend (approved):

svanherk (approved):
Thanks!

src/server/report/result.js
```diff
@@ -162,25 +153,32 @@ function renderResult(resultData, options) {
...
+			!resultData.passed && options.filterStatus === FILTER_STATUS.PASSED) {
+			return acc;
+		}
+
+		let status = STATUS_TYPE.WARNING;
+		if (resultData.duration > resultData.info.slowDuration) {
+			status = STATUS_TYPE.ERROR;
+		} else if (resultData.duration < (resultData.info.slowDuration / 2)) {
+			status = STATUS_TYPE.NORMAL;
+		}
```
dlockhart, line 169: Using these thresholds from Mocha's documentation:
Screenshot 2023-07-20 at 12 03 31 PM

src/server/report/result.js
dlockhart: I did have to add it to testInfo yeah. It would be cool if `@web/test-runner` exposed it to the reporters, but it does not (from what I can tell).

PR #81, opened 2023-07-19
handle case where golden doesn't exist yet

dlockhart:
Needed to re-organize the vdiff script a bit so that information about the actual test is provided before bailing.

Screenshot 2023-07-19 at 5 02 36 PM


src/server/visual-diff-plugin.js
```diff
@@ -191,36 +191,37 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 					path: updateGoldens ? goldenFileName : screenshotFileName
 				});
 
+				if (updateGoldens) {
+					return { pass: true };
+				}
+
 				const rootLength = join(rootDir, PATHS.VDIFF_ROOT).length + 1;
+
+				const screenshotImage = PNG.sync.read(await readFile(screenshotFileName));
```
dlockhart, line 200: Set information about the new screenshot early, before we bail if the golden is missing.

src/server/visual-diff-plugin.js
```diff
@@ -191,36 +191,37 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
-					return { pass: true };
-				}
-
 				const goldenExists = await checkFileExists(goldenFileName);
 				if (!goldenExists) {
+					setTestInfo(session, payload.name, {
+						new: {
+							path: `${screenshotFile.substring(rootLength)}.png`
+						}
+					});
```
dlockhart, line 215: If the golden isn't there, the "new" image is actually the "failed" one.

src/server/report/result.js
```diff
@@ -113,40 +114,43 @@ function renderResult(resultData, options) {
...
-				${img}
+				<div class="result-diff-container"><img src="../${partInfo.path}" loading="lazy" alt="">${overlay}</div>
 			</div>
 		`;
 	};
 
-	const overlay = (options.showOverlay && !resultData.passed) ?
+	const goldenExists = (resultData.info.golden !== undefined);
+
+	const overlay = (goldenExists && options.showOverlay && !resultData.passed) ?
```
dlockhart, line 132: Need to hide the overlay if there's no golden... since there won't be a diff to show.

bearfriend (approved):

PR #83, opened 2023-07-20
adding iterators

dlockhart:
You can now iterate between test files or tests within files. The "Home" button is now a breadcrumb on the left, and I made a few other minor stylistic tweaks.

Screenshot 2023-07-20 at 4 58 02 PM


src/server/report/app.js
```diff
@@ -160,20 +155,15 @@ class App extends LitElement {
...
+					<h1>Visual-diff Results</h1>
+					${this._renderFilters()}
+				</div>
+			</aside>
+			<main>${this._renderMainView()}</main>
 		`;
 	}
-	_handleBackClick() {
-		this._updateSearchParams({ file: undefined, test: undefined });
-	}
```
dlockhart, line 176: This and `_handleListTestClick` didn't end up being necessary due to the `<a href>` doing the same thing -- it was actually firing things twice.

src/server/report/app.js
```diff
@@ -312,6 +306,7 @@ class App extends LitElement {
 		return renderTabButtons('browser results', tabs, index => {
 			this._selectedBrowserIndex = index;
 			this.shadowRoot.querySelector('.tab-panels').scrollTo(0, 0);
+			this.requestUpdate();
```
dlockhart, line 309: `_selectedBrowserIndex` isn't a reactive state property anymore since we modify it ourselves. But in this case they've selected a new tab so we force a re-render.

src/server/report/app.js
```diff
@@ -349,9 +344,13 @@ class App extends LitElement {
...
 		});
 
-		const selectedBrowser = browsers[this._selectedBrowserIndex] ||
-			browsers.find(b => browserResults.get(b.name) < tests.length) ||
-			browsers[0];
+		if (!browsers[this._selectedBrowserIndex]) {
+			this._selectedBrowserIndex = Math.max(
+				browsers.findIndex((b) => browserResults.get(b.name) < tests.length),
+				0
+			);
```
dlockhart, line 351: We need to remember the index we automatically pick for them so that it stays as they iterate.

svanherk (approved):
Those chevrons are huge lol!

dlockhart:
> Those chevrons are huge lol!

Haha I know... they're our "tier 2" chevrons but could maybe use some slimming eh.

dlockhart:
Ok I've switched to an upscaled version of our tier1 chevrons instead, which are slimmer:
Screenshot 2023-07-24 at 4 22 08 PM

I'm sure Jeff will want to tweak these anyway, so don't want to spend too much time on them.

svanherk:
> I'm sure Jeff will want to tweak these anyway, so don't want to spend too much time on them.

Haha yeah I wasn't too worried about it for that reason, but they do look much better!

PR #85, opened 2023-07-21
Remove --manual option

bearfriend:


test/server/wtr-config.test.js
```diff
@@ -56,7 +56,7 @@ describe('WTRConfig', () => {
 		});
 
 		it('should set timeout to 0 and add headedMode plugin when headed', () => {
-			['watch', 'manual'].forEach(mode => {
+			['watch'].forEach(mode => {
```
bearfriend, line 59: This will be immediately followed up with adding `'open'` here, so leaving the `forEach`

dlockhart (approved):

PR #89, opened 2023-07-24
Add fixture pausing with UI controls

bearfriend:
[US152298](https://rally1.rallydev.com/#/?detail=/userstory/699970327679&fdp=true): Visual-diff: ability to debug in the browser

This should be the biggest PR to add the main feature. Everything after should be smaller additions and tweaks.

### DIFF
--- a/src/server/pause.js
+++ b/src/server/pause.js
@@ -51,8 +51,9 @@ const controls = `
 			cursor: pointer;
 		}
 
-		#d2l-test-controls button:focus {
-			background-color: #cdd5dc;
+		#d2l-test-controls button.primary {
+			color: #fff;
+			background-color: #006fbf;
 		}
 
 		#d2l-test-controls button:focus {
--- a/src/server/pause.js
+++ b/src/server/pause.js
@@ -68,6 +68,7 @@ const controls = `
 		}
 
 		#d2l-test-controls button[disabled] {
+			cursor: default;
 			opacity: 0.5;
 		}
 	</style>
--- a/src/server/pause.js
+++ b/src/server/pause.js
@@ -1,0 +1,131 @@
+const test = window.d2lTest = {};
+test.pause = new Promise(r => test.start = r);
+
+const controls = `
+       <style>
+               #d2l-test-controls [hidden] {
+                       display: none !important;
+               }
+
+               #d2l-test-controls #start, #run {
+                       display: flex;
+                       gap: 1em;
+                       font-size: 18px;
+                       font-family: 'Lato', sans-serif;
+                       padding: 1em;
+                       background: #fff;
+                       color: #222;
+                       margin: calc(-30px + -8px) calc(-30px + -8px) calc(30px + 8px);
+                       flex-wrap: wrap;
+                       align-items: center;
+                       box-shadow: 0 0 5px rgba(0,0,0,.5);
+               }
+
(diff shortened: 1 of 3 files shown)


src/server/pause.js
```diff
@@ -0,0 +1,131 @@
...
+			cursor: pointer;
+		}
+
+		#d2l-test-controls button:focus {
+			background-color: #cdd5dc;
+		}
+
+		#d2l-test-controls button:focus {
+			background-color: #cdd5dc;
+		}
```
dlockhart, line 60: Looks duplicated from lines 54-56?

src/server/pause.js
```diff
@@ -0,0 +1,131 @@
+const test = window.d2lTest = {};
+test.pause = new Promise(r => test.start = r);
+
+const controls = `
+	<style>
+		#d2l-test-controls [hidden] {
+			display: none !important;
+		}
+
+		#d2l-test-controls #start, #run {
```
dlockhart, line 10: Since these end up in the global namespace, I might suggest prefixing everything with `d2l-test-*` (i.e. `#d2l-test-run`, `#d2l-test-start`) just to avoid the crazy situation where someone uses an `id="start"` on one of their top-level components.

src/server/pause.js
```diff
@@ -0,0 +1,131 @@
...
+		#d2l-test-controls button:focus {
+			background-color: #cdd5dc;
+		}
+
+		#d2l-test-controls button:focus-visible {
+			box-shadow: 0 0 0 2px #fff, 0 0 0 4px #006fbf;
+		}
+
+		#d2l-test-controls button.primary:focus {
+			background-color: #004489;
```
dlockhart, line 67: Was this intentional?
Screenshot 2023-07-24 at 2 44 46 PM

Suggest either always having the background color set on "primary" buttons and also setting the foreground to white, or just not having the concept of primary.

src/server/pause.js
```diff
@@ -0,0 +1,131 @@
...
+		}
+
+		#d2l-test-controls button[disabled] {
+			opacity: 0.5;
```
dlockhart, line 71:
```suggestion
			cursor: default;
			opacity: 0.5;
```

dlockhart (commented):
Looking nice!

src/server/pause.js
dlockhart: Ah I see this was resolved in a follow-up PR.

bearfriend: Yeah things are slightly messy trying to piece it all back together from the original big PR.

src/server/pause.js
bearfriend: I guess you mean to prevent them from accidentally selecting the controls? Not entirely opposed to that, but I think I'd rather convert the whole thing to a component after we get it merged.

dlockhart: Yeah either a `getElementById` type collision or in this case this CSS would target their stuff and do weird things. But if the plan is to wrap it in a shadow root, then those concerns go away.

bearfriend: Well this targets stuff inside `#d2l-test-controls` (except I missed `#run` here), but point taken.

dlockhart (approved):
Looks good -- just the duplicated CSS note, but maybe that'll get cleaned up later.

PR #92, opened 2023-07-24
Maintain LTR controls when RTL

bearfriend:


dlockhart (approved):
Haha yeah weird that we'd ever need to do this, but in this case it makes sense!

svanherk (approved):

PR #93, opened 2023-07-24
Fix headed Chrome screenshot

bearfriend:
Fixes headed Chrome screenshots by inlining all styles to capture hover effects before they break.


src/browser/commands.js
```diff
@@ -30,6 +30,7 @@ export async function focusElem(elem) {
 
 export async function hoverAt(x, y) {
 	await sendMouse({ type: 'move', position: [x, y] });
+	if (window.d2lTest) window.d2lTest.hovering = true;
```
dlockhart, line 33: Would having the vdiff/pause stuff just add a listener on the window or body for `mouseover` also work? If so, it might be nice for that code to live over there where it belongs.

dlockhart (approved):
Crazy that we need to do this, but also crazy that it works!

bearfriend:
> Crazy that we need to do this, but also crazy that it works!

Yeah, I'd definitely like to get rid of it sooner rather than later if possible.

PR #94, opened 2023-07-24
Surface browser version change information

dlockhart:
This adds a tracker `.vdiff.json` file to the root of the repo-under-test that is intended to get checked into source control. In there we can put whatever we like, but for now it contains the browsers and versions of those browsers from the previous test run.

We can then use that information to surface any changes in the versions in the report -- both in the side panel but also when viewing a particular browser's results:

Screenshot 2023-07-24 at 3 13 26 PM

### DIFF
--- a/.vdiff.json
+++ b/.vdiff.json
@@ -1,0 +1,16 @@
+{
+	"browsers": [
+		{
+			"name": "Chromium",
+			"version": 115
+		},
+		{
+			"name": "Firefox",
+			"version": 115
+		},
+		{
+			"name": "Webkit",
+			"version": 17
+		}
+	]
+}
--- a/src/server/visual-diff-reporter.js
+++ b/src/server/visual-diff-reporter.js
@@ -25,7 +25,7 @@ function createData(rootDir, sessions) {
 			browsers.set(browserName, {
 				name: browserName,
 				numFailed: 0,
-				version: parseInt(s.browser.browser.version().substring(0, s.browser.browser.version().indexOf('.'))),
+				version: parseInt(s.browser.browser.version()),
 				previousVersion: prevBrowser?.version
 			});
 		}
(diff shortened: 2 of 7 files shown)


.vdiff.json
dlockhart, line 1: This is what the file we'll be checking into each repo would look like. Open to other suggestions around naming, etc.

src/server/visual-diff-reporter.js
```diff
@@ -8,17 +8,25 @@ const __dirname = dirname(fileURLToPath(import.meta.url));
...
 	sessions.forEach(s => {
 
 		const browserName = s.browser.name;
 		if (!browsers.has(browserName)) {
+			const prevBrowser = metadata.browsers?.find(b => b.name === browserName);
 			browsers.set(browserName, {
 				name: browserName,
 				numFailed: 0,
-				version: s.browser.browser.version().substring(0, s.browser.browser.version().indexOf('.'))
+				version: parseInt(s.browser.browser.version().substring(0, s.browser.browser.version().indexOf('.'))),
```
bearfriend, line 28: I think `parseInt` is going to stop at the `.` anyway.
```suggestion
				version: parseInt(s.browser.browser.version()),
```

dlockhart: That's a nice side benefit, thanks!

.vdiff.json
dlockhart, line 1: This is what the file we'll be checking into each repo would look like. Open to other suggestions around naming, etc.

bearfriend (approved):

svanherk:
Sorry, somehow totally missed this PR. So when do we commit the updates to this file?  Only when tests have failed?

dlockhart:
> So when do we commit the updates to this file?

I was hoping it would get committed as part of the visual-diff action just like a change to a golden?

svanherk:
Locally, I guess we never want to update it because we don't want people accidentally committing it?

Then in IC, if the tests passed the action isn't currently set up to commit anything (even orphaned goldens) because those would just be handled next time there's an actual failure or update needed.  But in this case, it's a bit more awkward - you could run into a scenario where the goldens have passed for a few weeks and the next failure jumps up a few browser versions.  In practice, maybe this would never ever happen - a new browser version always causes a failure.

And technically, the data is still accurate - the previous goldens were generated at that browser version from weeks ago.

So maybe what we want is?

- If running locally, have a copy of the config in `.vdiff` that's created with `--golden` and updated during runs, but not committed (handled by the `.gitignore`)?
- If running in CI, always create/update/use the .vdiff.json` copy, and if any goldens _are_ updated it will be committed alongside it?

PR #97, opened 2023-07-25
Fix Windows headed mode import

bearfriend:
The `pausePath` is for an import path, not a filesystem path, so always use POSIX-style forward slash. Also, remove leading `/C:`.

Using an absolute path to the file works because the dev server will resolve and transform it as appropriate for the OS.


dlockhart (approved):

svanherk (approved):
It's alive!!!

PR #101, opened 2023-07-26
expose "slow" CLI parameter and pass-through to Mocha

dlockhart:
Mocha has a concept of a "slow" threshold that it uses to highlight slow/normal/fast tests in reports. It defaults to 75ms. Tests which run in at most half that value are "fast", tests which take longer than that value are "slow" and everything else is "normal".

This exposes `slow` as a CLI argument to `d2l-test-runner` and passes it through to Mocha (similar to `timeout`). It also sets a default value for vdiff tests of `500ms`.

### DIFF
--- a/src/server/cli/test-runner.js
+++ b/src/server/cli/test-runner.js
@@ -1,9 +1,9 @@
 import { ConfigLoaderError, readConfig } from '@web/config-loader';
+import { DEFAULT_VDIFF_SLOW, WTRConfig } from '../wtr-config.js';
 import commandLineArgs from 'command-line-args';
 import commandLineUsage from 'command-line-usage';
 import process from 'node:process';
 import { startTestRunner } from '@web/test-runner';
-import { WTRConfig } from '../wtr-config.js';
 
 async function getTestRunnerOptions(argv = []) {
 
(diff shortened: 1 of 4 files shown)


src/server/wtr-config.js
```diff
@@ -14,6 +14,7 @@ const BROWSER_MAP = {
 	safari: 'webkit',
 	webkit: 'webkit'
 };
+const DEFAULT_VDIFF_SLOW = 500;
```
dlockhart, line 17: I'm guessing we'll want to tweak this later once we have vdiff running in core and have a better idea of what "slow" means for visual-diff. We may also want different values locally compared to CI.

src/server/cli/test-runner.js
```diff
@@ -96,6 +96,13 @@ async function getTestRunnerOptions(argv = []) {
 			longAlias: 'webkit',
 			type: Boolean
 		},
+		{
+			name: 'slow',
+			alias: 's',
+			type: Number,
+			description: 'Tests whose duration in milliseconds are at most half of this threshold are "fast" and tests which exceed it are "slow"\n[Default: 75]',
```
svanherk, line 103: Maybe worth calling out both defaults (500 for vdiff, 75 for everyone else)?

dlockhart: Cool -- done.

svanherk (approved):

PR #108, opened 2023-07-31
feat: support configuring mathjax context

dlockhart:
This is meant to remove the need for [this plugin](https://github.com/BrightspaceUI/core/tree/main/components/html-block#add-context-automatically-to-demos-and-tests) which includes [this script](https://github.com/BrightspaceUI/core/blob/main/tools/mathjax-test-context.js) in unit and visual-diff tests.

Consumers will be able to enable LaTeX in their tests by passing `{ mathjax: { renderLatex: true } }` as an option to `fixture()`. That'll get reset in between each test. If required, I'm leaving the door open to more mathjax configuration options.

### DIFF
--- a/src/browser/reset.js
+++ b/src/browser/reset.js
@@ -2,10 +2,12 @@ import { sendMouse, setViewport } from '@web/test-runner-commands';
 import { nextFrame } from '@open-wc/testing';
 
 const DEFAULT_LANG = 'en',
+	DEFAULT_MATHJAX_RENDER_LATEX = false,
 	DEFAULT_VIEWPORT_HEIGHT = 800,
 	DEFAULT_VIEWPORT_WIDTH = 800;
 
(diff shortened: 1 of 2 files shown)


dlockhart:
@devpow112 I see you spotted this -- interested in your thoughts too!

grant-cleary (approved):

dbatiste (approved):

devpow112:
> @devpow112 I see you spotted this -- interested in your thoughts too!

I'm definitely happy to make anything simpler. The script was definitely just a workaround (mainly for rubrics/outcomes) when doing test/vdiff stuff (vdiff didn't seem to like me just setting the mathjax context in the HTML directly). Having it be part of the framework makes me happy 😄

devpow112 (approved):

svanherk (approved):

PR #113, opened 2023-08-02
Add vdiff subcommands

bearfriend:
- Moves `migrate-goldens` to `d2l-test-runner vdiff migrate`
- Moves `d2l-test-vdiff-report` to `d2l-test-runner vdiff report`
- Moves `--golden` option to a subcommand `d2l-test-runner vdiff golden`
- Adds subcommands to `--help` output
- Adds clarifying output before running tests to generate goldens
- Adds `dtr` shortcut command


src/server/cli/vdiff/migrate.js
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env node
+import { appendFile, mkdir, readFile, rename, rm } from 'node:fs/promises';
+import { join, normalize, parse } from 'node:path';
+import commandLineArgs from 'command-line-args';
+import { glob } from 'glob';
+import { PATHS } from '../../visual-diff-plugin.js';
+import { stdout } from 'node:process';
+
+export async function migrate(argv = []) {
+	const { pattern = './**' } = commandLineArgs({ name: 'pattern', type: String, defaultOption: true }, { partial: true, argv });
```
bearfriend, line 10: Only changes to this from the `bin` version were wrapping it in a function and changing the default pattern from `./test/**` to `./**` because there's no reason to restrict where to look for goldens by default.

svanherk: Will this look in `node_modules`?  I remember that coming up with the old library

bearfriend: It will, but it's being ignored on line 12

bin/d2l-test-runner.js
```diff
@@ -1,7 +1,33 @@
...
+
+	if (!vdiff.subcommand) {
+		runTests();
+	} else if (vdiff.subcommand === 'golden') {
+		argv.splice(argv.findIndex(a => a === 'golden'), 1, '--golden');
+		stdout.write('\nGenerating vdiff goldens...\n');
+		runTests();
+	} else if (vdiff.subcommand === 'report') {
+		import('../src/server/cli/vdiff/report.js');
+	}	else if (vdiff.subcommand === 'migrate') {
```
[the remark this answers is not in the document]
svanherk: Is our linting not setup properly, or does the `browser-config` not care about extra spaces?

package.json
```diff
@@ -15,8 +15,7 @@
   },
   "bin": {
     "d2l-test-runner": "./bin/d2l-test-runner.js",
-    "d2l-test-vdiff-report": "./bin/d2l-test-vdiff-report.js",
-    "migrate-goldens": "./bin/migrate-goldens.js"
+    "dtr": "./bin/d2l-test-runner.js"
```
svanherk, line 18: Lol not sure I'll be able to use this without reading it as "define the relationship"

src/server/bail.js
bearfriend, line 1: So, all this `bail` stuff is just sort of an experiment. I don't love having specific test code in `report.js`, though it's extremely minimal. It could easily be set up the same way as `migrate` to export a `start` method that can be stubbed.

test/bin/d2l-test-runner.test.js
```diff
@@ -1,18 +1,79 @@
 import { assert, restore, stub } from 'sinon';
-import { argv } from 'node:process';
+import { bail } from '../../src/server/bail.js';
+import { expect } from 'chai';
+import { migrate } from '../../src/server/cli/vdiff/migrate.js';
+import process from 'node:process';
 import { runner } from '../../src/server/cli/test-runner.js';
 
+const { argv, stdout } = process;
```
bearfriend, line 8: Importing `process` and destructuring `argv` allows us to modify it globally.

src/server/bail.js
dlockhart: Yeah definitely feel free to modify it to behave more like `migrate` if that's easier!

bearfriend: I think I will because it works with the pattern but it's good to know this way works too.

PR #115, opened 2023-08-03
initial pass at documenting "writing tests"

dlockhart:
I'm going to do this in a few phases for my and your sanity. This one reorganizes the general structure of the README and includes docs for "writing tests".

There's a fair bit of repetition from what's in the `@open-wc` documentation, but I feel like it's worth it since we adapted most of their APIs. Also not having to bounce around is nice.

Next up will be "running tests" which will go into the binaries/CLI aspect of things. After that, the migration guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 [![NPM version](https://img.shields.io/npm/v/@brightspace-ui/testing.svg)](https://www.npmjs.org/package/@brightspace-ui/testing)
 
-Testing utilities that build upon [@web/test-runner](https://modern-web.dev/docs/test-runner/overview/) but are specifically designed and configured to work with Brightspace components and applications.
+Testing utilities which are specifically designed and configured for Brightspace UI components and applications.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,7 +2,7 @@
 
 [![NPM version](https://img.shields.io/npm/v/@brightspace-ui/testing.svg)](https://www.npmjs.org/package/@brightspace-ui/testing)
 
-Testing utilities that build upon [@web/test-runner](https://modern-web.dev/docs/test-runner/overview/) but are specifically designed and configured to work with Brightspace components and applications.
+Testing utilities which are specifically designed and configured for Brightspace UI components and applications.
```
dlockhart, line 5: I don't want to take away that this is building on other stuff, but I moved this information down to where it's most relevant.

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
+  }
+);
+
+it('should use custom element', async() => {
+  const foo = 'bar';
+  const elem = await fixture(html`<${tag} foo="${foo}"></${tag}>`);
+});
 ```
 
-## Versioning & Releasing
```
dlockhart, line 36: The diff got pretty confused, but I've trimmed down our "semantic-release" blurb significantly since it repeats so much information that's already in the action README. If folks like this, I'd like to make it the template for all repos.

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
+Note that the call to `clickElem` is not `await`-ed, since by the time it resolves the event will likely have already been dispatched.
+
+An even safer approach would be to wrap that call in a `setTimeout`:
+
+```javascript
+setTimeout(() => clickElem(elem));
+await oneEvent(elem, 'some-event');
+```
+
+#### Waiting for a Lit Element to Update
```
dlockhart, line 204: This could easily be left out, but it could be useful to call it out in case folks use `aTimeout` or `nextFrame` instead of realizing they can do this? The [open-wc stuff](https://open-wc.org/docs/testing/helpers/#elementupdated) has a special `await elementUpdated(elem)` that essentially does this and falls back to `nextFrame` if it's not a Lit element.

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
+it('should wait for an event', async() => {
+  const elem = await fixture(html`<my-elem></my-elem>`);
+  clickElem(elem);
+  await oneEvent(elem, 'some-event');
+});
+```
 
-# unit tests only
-npm run test:headless
+Note that the call to `clickElem` is not `await`-ed, since by the time it resolves the event will likely have already been dispatched.
```
bearfriend, line 195: I feel like it's clearer to me to state it the other way around.
```suggestion
Note that the call to `clickElem` is not `await`-ed, since by the time it resolves the event will have already been dispatched.
```
```suggestion
Note that the call to `clickElem` is not `await`-ed, so that the listener is registered before the event is dispatched.
```
Fine either way though. Whatever people prefer.

margaree (commented):
I can't completely evaluate this for accuracy of documenting new functionality (though I think that all seems fine), but it was a useful read for me to see what's changed and overall I found the documentation clear.

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
+An even safer approach would be to wrap that call in a `setTimeout`:
+
+```javascript
+setTimeout(() => clickElem(elem));
+await oneEvent(elem, 'some-event');
+```
+
+#### Waiting for a Lit Element to Update
+
+When using Lit-based components, it's common to create a fixture, modify some properties and then want to wait for those changes to be rendered.
```
bearfriend, line 206: Just seems clearer to me
```suggestion
When using Lit-based components, it's common to create a fixture and then modify some of its properties, which usually requires waiting for those changes to be rendered.
```

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
 
-> TL;DR: Commits prefixed with `fix:` and `feat:` will trigger patch and minor releases when merged to `main`. Read on for more details...
+Unlike other tests, visual-diff tests must be in a file with the `*.vdiff.js` extension.
 
-The [semantic-release GitHub Action](https://github.com/BrightspaceUI/actions/tree/main/semantic-release) is called from the `release.yml` GitHub Action workflow to handle version changes and releasing.
+Use the asynchronous `.to.be.golden()` Chai assertion to take a visual-diff snapshot and compare it against its golden.
 
-### Version Changes
+```javascript
+import { fixture, html } from '@brightspace-ui/testing';
```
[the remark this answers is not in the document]
dlockhart: _Possibly_, yeah. Although really visual-diff is just getting in line with our other "unit" tests which always had to do those imports. I'll think about this a bit more.

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
+);
+
+it('should use custom element', async() => {
+  const foo = 'bar';
+  const elem = await fixture(html`<${tag} foo="${foo}"></${tag}>`);
+});
 ```
 
-## Versioning & Releasing
+> **Important:** `defineCE` is not performant and shouldn't be used outside of tests.
```
bearfriend, line 308:
```suggestion
> **Important:** `defineCE` is not performant and shouldn't be used outside of test files.
```

README.md
```diff
@@ -12,62 +12,370 @@ Install from NPM:
...
+
+### Visual-Diff Testing
+
+Also known as "visual regression" or "perceptual diff", visual-diff testing involves taking snapshot images of the browser and comparing them against a known golden (or "baseline") image. The images are compared pixel-by-pixel and differences beyond a threshold will fail the test. Our visual diff testing leverages the [pixelmatch](https://github.com/mapbox/pixelmatch) library to perfom its comparison.
 
-> TL;DR: Commits prefixed with `fix:` and `feat:` will trigger patch and minor releases when merged to `main`. Read on for more details...
+Unlike other tests, visual-diff tests must be in a file with the `*.vdiff.js` extension.
```
bearfriend, line 314:
```suggestion
Unlike other tests, visual-diff tests must be in a file that matches the pattern `*.vdiff.js`.
```

README.md
bearfriend: Oh, or does this mean used the result of `defineCE` in tests?

README.md
bearfriend: Actually, this isn't quite accurate either since they can defines their own pattern. It just uses `vdiff` as the type _in_ that pattern.

README.md
dlockhart: From the open-wc docs:
> This uses a workaround that's not performant for re-renders, which is fine for testing, but do NOT use this in production code.

README.md
dlockhart: Right. So should I just leave this part totally off and it can be discussed down in the "running tests" part?

README.md
bearfriend: Ok, then I still with the suggestion above for clarity since `defineCE` is usually used outside of the literal `it` test, but always only in test files.

README.md
bearfriend: Yeah maybe that's best

bearfriend (approved):

margaree (approved):

README.md
```diff
@@ -12,62 +12,361 @@ Install from NPM:
...
 
-### Linting
+### describe, it, assert & expect
 
-```shell
-# eslint
-npm run lint
+Tests can be grouped into suites using `describe` and are defined inside `it` blocks.
+
+Results can be verified using either the BDD-style `expect` or TDD-style `assert` syntax (although try not to mix & match).
```
svanherk, line 23: > (although try not to mix & match)

Lol and then you do just that. Maybe broken into two different code blocks?
```javascript
import { expect } from '@brightspace-ui/testing';

describe('group of tests', () => {
  it('should multiply numbers', () => {
    expect(2 * 4).to.equal(8);
  });
});
```
```javascript
import { assert} from '@brightspace-ui/testing';

describe('group of tests', () => {
  it('should multiply numbers', () => {
    assert.equal(2 * 4, 8);
  });
});
```

Maybe that's overkill

README.md
```diff
@@ -12,62 +12,361 @@ Install from NPM:
...
+  const elem = await fixture(html`<my-elem></my-elem>`, {
+    rtl: true
+  });
+  // do assertions
+});
+```
+
+If `lang` is set to Arabic (`ar`), the right-to-left option will automatically be enabled.
+
+> **Note:** it's not recommended to use `language` configuration with visual-diff to solely test the correctness of translations. The [messageformat-validator](https://github.com/bearfriend/messageformat-validator) is a more efficient way to test translations.
```
svanherk, line 88: Lol we should probably get that moved into `BrightspaceUI` at some point

README.md
```diff
@@ -12,62 +12,361 @@ Install from NPM:
...
+import { fixture, focusElem } from '@brightspace-ui/testing';
+
+it('should focus on element', async() => {
+  const elem = await fixture(html`<button>elem</button>`);
+  await focusElem(elem);
+  // do assertions
+});
+```
+
+To send particular keystrokes to the browser window or a specific element, use `sendKeys(action, keys)` or `sendKeysElem(elem, action, keys)`.
```
svanherk, line 154: I think it'd be worth calling out somewhere in here that `sendKeysElem` will focus the element first, not just fire a keyboard event from it

README.md
```diff
@@ -12,62 +12,361 @@ Install from NPM:
...
+  const foo = 'bar';
+  const elem = await fixture(html`<${tag} foo="${foo}"></${tag}>`);
+});
+```
+
+> **Important:** `defineCE` is not performant and shouldn't be used outside of test files.
+
+### Visual-Diff Testing
+
+Also known as "visual regression" or "perceptual diff", visual-diff testing involves taking snapshot images of the browser and comparing them against a known golden (or "baseline") image. The images are compared pixel-by-pixel and differences beyond a threshold will fail the test. Our visual diff testing leverages the [pixelmatch](https://github.com/mapbox/pixelmatch) library to perfom its comparison.
```
svanherk, line 305: Maybe worth calling it `vdiff` in at least a few places here, or everywhere?  That's what we've been using everywhere else, in the command, action, test names, etc.

README.md
```diff
@@ -12,62 +12,361 @@ Install from NPM:
...
+  describe('situation1', () => {
+    it('state1', async() => {
+      const elem = await fixture(html`<my-elem></my-elem>`);
+      await expect(elem).to.be.golden();
+    });
+  });
+});
+```
+
+The filename and location of the resulting image will be based on the suite names and test name. The top-most suite will become a directory, and any remaining suites will be combined with the test name into the file name. In this example, the snapshot would be stored in: `./my-elem/situation1-state1.png`
```
svanherk, line 322: Are we going to have a separate section with more info on `vdiff`?  This path isn't really accurate, for local or for ci, and I figured we'd talk about what file names are expected, link to the action, etc.

README.md
```diff
@@ -12,62 +12,361 @@ Install from NPM:
...
+
+# vdiff tests
+npm run test:vdiff
+```
 
-Occasionally you'll want to backport a feature or bug fix to an older release. `semantic-release` refers to these as [maintenance branches](https://semantic-release.gitbook.io/semantic-release/usage/workflow-configuration#maintenance-branches).
+### Versioning & Releasing
 
-Maintenance branch names should be of the form: `+([0-9])?(.{+([0-9]),x}).x`.
+This repo is configured to use the `semantic-release`. Commits prefixed with `fix:` and `feat:` will trigger patch and minor releases when merged to `main`.
```
svanherk, line 370:
```suggestion
This repo is configured to use the `semantic-release` action. Commits prefixed with `fix:` and `feat:` will trigger patch and minor releases when merged to `main`.
```
```suggestion
This repo is configured to use `semantic-release`. Commits prefixed with `fix:` and `feat:` will trigger patch and minor releases when merged to `main`.
```

svanherk (commented):
This was merged when I was halfway through my review lol, but most were minor things anyways! Figured it was still worth posting though

README.md
dlockhart: Yeah, or even into this library! `d2l-test-mfv`

PR #117, opened 2023-08-04
Ignore - Add some debug logs

svanherk:
Try checking if it's the mouse move problem on browser startup again


svanherk:
Haven't really been seeing this since? Closing, we'll come back to this when it pops up again

PR #118, opened 2023-08-04
minor tweaks

dlockhart:
Based on @svanherk's feedback on #115.


README.md
```diff
@@ -319,7 +322,7 @@ describe('my-elem', () => {
 });
 ```
 
-The filename and location of the resulting image will be based on the suite names and test name. The top-most suite will become a directory, and any remaining suites will be combined with the test name into the file name. In this example, the snapshot would be stored in: `./my-elem/situation1-state1.png`
+The filename and location of the resulting image will be based on the suite names and test name.
```
dlockhart, line 325: I'll probably include more on filenames somewhere else, so removing this for now.

PR #119, opened 2023-08-04
Fix golden flag check

bearfriend:
Importing just `argv` makes it a static reference to the original `argv`, so transforming the `golden` command into a `--golden` option on `process.argv` isn't picked up.


dlockhart (approved):
👍

PR #125, opened 2023-08-10
remove installation of playwright deps

dlockhart:
Hopefully we don't need this. 🤞


dlockhart:
Apparently we do, :(.

PR #126, opened 2023-08-10
feat: install playwright dependencies ourselves

dlockhart:
To avoid consumers having to run this in CI anytime they use `d2l-test-runner`, this installs it automatically for them. In testing, it appears to be a no-op if they're already installed.

### DIFF
--- a/bin/d2l-test-runner.js
+++ b/bin/d2l-test-runner.js
@@ -29,6 +29,7 @@ if (cli.subcommand === 'vdiff') {
 }
 
 async function runTests() {
+	await runner.install();
 	const options = await runner.getOptions(argv);
 	await runner.start(options);
 }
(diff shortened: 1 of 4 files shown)


src/server/cli/test-runner.js
```diff
@@ -196,7 +197,12 @@ async function getTestRunnerOptions(argv = []) {
 	};
 }
 
+function installDeps() {
+	execSync('npx playwright install-deps', { stdio: 'pipe' });
```
dlockhart, line 201: I played around with the async version of this and wrapped it in a Promise, but marshalling the stdout/err around seemed way more complicated than it was worth. Open to other suggestions though.

bin/d2l-test-runner.js
```diff
@@ -29,6 +29,7 @@ if (cli.subcommand === 'vdiff') {
 }
 
 async function runTests() {
+	await runner.install();
```
bearfriend, line 32:
```suggestion
	runner.install();
```

bearfriend (approved):

PR #127, opened 2023-08-10
US156332 - Handle passing document.body to get a full page screenshot

svanherk:
Visual diff test result looks like this (quoted to help show where it begins and ends lol): 
> image

### DIFF
--- a/src/browser/vdiff.js
+++ b/src/browser/vdiff.js
@@ -27,7 +27,7 @@ async function ScreenshotAndCompare(opts) {
 	}
 
 	const name = this.test.fullTitle();
-	const rect = this.elem.getBoundingClientRect();
+	const rect = this.elem === document.body ? null : this.elem.getBoundingClientRect();
 	const slowDuration = this.test.slow();
 	let result = await executeServerCommand('brightspace-visual-diff-compare', { name, rect, slowDuration, opts });
 	if (result.resizeRequired) {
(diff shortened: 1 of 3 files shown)


dlockhart (approved):
Awesome!

src/browser/vdiff.js
```diff
@@ -27,7 +27,7 @@ async function ScreenshotAndCompare(opts) {
 	}
 
 	const name = this.test.fullTitle();
-	const rect = this.elem.getBoundingClientRect();
+	const rect = this.elem === document.body ? null : this.elem.getBoundingClientRect();
```
bearfriend, line 30: Is the `html` tag technically a better match for what we capture here? It's just capturing the viewport, right?
```suggestion
	const rect = this.elem === document.documentElement ? null : this.elem.getBoundingClientRect();
```

svanherk: Either works! `documentElement ` is probably more accurate, but `body` easier to remember? I don't have a strong feeling, we could literally pass in 'page' as a string lol

svanherk: Ooo, I like that

README.md
```diff
@@ -335,6 +335,21 @@ describe('my-elem', () => {
 
 The filename and location of the resulting image will be based on the suite names and test name.
 
+Sometimes you might be writing tests to verify a component's position within the whole viewport. For snapshots of the whole page, set up your components and pass `document` to the assertion:
```
svanherk, line 338: I don't love this, feel free to wordsmith it.  But I did use the right form of set up!

dlockhart: I'm going to rewrite all this anyway to separate vdiff out into its own section, so can have another look then.

dlockhart (approved):

bearfriend (approved):

PR #128, opened 2023-08-10
fix: exclude skipped tests from report

dlockhart:
Tests which were skipped (e.g. via `grep`) were showing up in the report but didn't have any "info". This excludes them.


dlockhart:
Actually this isn't quite right, closing for now.

PR #144, opened 2023-08-23
Add `migrate-local` subcommand

bearfriend:


bin/d2l-test-runner.js
```diff
@@ -21,8 +22,15 @@ if (cli.subcommand === 'vdiff') {
 	} else if (vdiff.subcommand === 'migrate') {
 		const { migrate } = await import('../src/server/cli/vdiff/migrate.js');
 		await migrate.start(vdiff._unknown);
+	} else if (vdiff.subcommand === 'migrate-local') {
+
+		execSync('npm install @brightspace-ui/visual-diff@14  --no-save');
+		execSync('npx mocha \'./**/*.visual-diff.js\' -t 10000 --golden');
```
svanherk, line 28: Probably not a big deal in smaller repos, but in `core` this takes forever.  Would it work to accept a folder name (like `dialog`) and then do this:
```
execSync('npx mocha \'./**/dialog/*.visual-diff.js\' -t 10000 --golden');
```
and then in `migrate` do:
```
const dirs = await glob(`./**/${pattern}/*/${oldSuffix}`, { ignore: 'node_modules/**', posix: true });
```
?
Would solve the problem I have of having to run it ten times before I find the write pattern format

src/server/cli/vdiff/migrate.js
```diff
@@ -30,7 +30,9 @@ async function start(argv = []) {
 				.replace(/^d2l-/, '')
 				.replace(new RegExp(`^${dirName}-`), '');
 
-			const newDir = dir.replace(`${oldSuffix}/${dirName}`, `${PATHS.GOLDEN}/${dirName}/chromium`);
+			const newDir = local ?
+				`${PATHS.VDIFF_ROOT}/${dir.replace(`${oldSuffix}/${dirName}`, `${dirName}/${PATHS.GOLDEN}/chromium`)}` :
+				dir.replace(`${oldSuffix}/${dirName}`, `${PATHS.GOLDEN}/${dirName}/chromium`);
 
 			await mkdir(newDir, { recursive: true });
 			return rename(file, join(newDir, newName));
```
svanherk, line 38: I think the migration is working, but then line 40 below deletes the whole test folder that `screenshots` was in?

svanherk (approved):
Seems to be working for me!

svanherk:
Oh, probably need to update the printed docs somewhere

PR #146, opened 2023-08-24
US156331 - Add ability to indicate the actual element to use for size calculations for the screenshot

svanherk:
Some root custom elements that consumers pass to vdiff won't be the one that defines its true size. For example, the `<d2l-dialog>` element has no height or width -- instead, the size is defined by an element in its shadow root. As a result, snapshotting a dialog fails.

This PR adds functionality to look for the "real" size. If your custom element is one of these, you can add the ~~`vdiff-size`~~ `vdiff-target` class to the element in its `shadowRoot` that _does_ have the proper size. This will search recursively, until there's no element in the `shadowRoot` with ~~`vdiff-size`~~ `vdiff-target` or we reach a native HTML element. I've added a couple tests to demonstrate this.

Not tied to _any_ of the naming, so suggest away!

Without the changes to check for ~~`vdiff-size`~~ `vdiff-target`, the new goldens here would look like this: 
image


test/browser/element.vdiff.js
```diff
@@ -39,6 +40,40 @@ const elementTag = defineCE(
...
+					border-radius: 8px;
+					box-shadow: 0 0 10px rgba(0, 0, 0, 0.1);
+					box-sizing: border-box;
+					position: fixed;
+				}
+			`;
+		}
+		render() {
+			return html`
+				<div class="vdiff-size">
```
svanherk, line 59: This represents typical `core` usage. We'll add the appropriate class to `dialog`, dropdown`, etc.

test/browser/element.vdiff.js
```diff
@@ -39,6 +40,40 @@ const elementTag = defineCE(
...
+			`;
+		}
+	}
+);
+
+const nestedElementTag = defineCE(
+	class extends LitElement {
+		render() {
+			return html`${unsafeHTML(`
+				<${fixedElementTag} class="vdiff-size"><${elementTag} text="Visual Difference"></${elementTag}></${fixedElementTag}>
```
svanherk, line 71: This represents what is likely to happen for consumers of `core`.  If they have a component that wraps `d2l-dialog` for example, they'll add the class to `d2l-dialog`.  vdiff will look for this class, then follow it into `d2l-dialog` to use the element we've added the class to.

src/server/wtr-config.js
```diff
@@ -196,6 +196,7 @@ export class WTRConfig {
 		config.groups.forEach(g => g.browsers = this.getBrowsers(g.browsers));
 
 		if (open || watch) {
+			config.testsFinishTimeout = 15 * 60 * 1000;
```
svanherk, line 199: This is a bit unrelated, but when I was trying to debug my new tests with `--open`, I only had two minutes before the whole thing force-closed.  This ups that timeout to 15 minutes in `open` or `watch` mode.

src/browser/vdiff.js
```diff
@@ -19,6 +19,13 @@ mocha.setup({
 });
 /* eslint-enable */
 
+function findRealSize(elem) {
+	if (!elem.shadowRoot) return elem;
+	const nestedSizedElem = elem.shadowRoot.querySelector('.vdiff-size');
+	if (!nestedSizedElem) return elem;
+	return findRealSize(nestedSizedElem);
```
dlockhart, line 26: Cool. So if an element doesn't define its size at all but instead that falls to something in its shadow root, it can just put the class on that element and it'll recurse -- nice!

src/browser/vdiff.js
```diff
@@ -19,6 +19,13 @@ mocha.setup({
 });
 /* eslint-enable */
 
+function findRealSize(elem) {
+	if (!elem.shadowRoot) return elem;
+	const nestedSizedElem = elem.shadowRoot.querySelector('.vdiff-size');
```
dlockhart, line 24: Alternate options for the CSS class: `vdiff-rect-elem`, `vdiff-rect`, `vdiff-container`

dlockhart (approved):
Seems straightforward enough! Not sure about the CSS class name, but offered a few suggestions.

src/browser/vdiff.js
svanherk: Yeah exactly! And in theory, this should "just work" for consumers wrapping a `core` component in that situation, because we would've had to already add it for everything that needs it for our own tests.

src/browser/vdiff.js
svanherk: Yeah I had `vdiff-rect` and then I didn't love it, but I don't love `vdiff-size` either so switched back to that!

dlockhart: Throwing another option out there: `vdiff-boundary`

svanherk: `vdiff-area`? `vdiff-box`? `vdiff-frame`? `vdiff-actual`?

bearfriend: `vdiff-target`?

PR #151, opened 2023-08-28
Expose oneDefaultPreventedEvent from open-wc/testing

svanherk:
I need this in the new dialog unit test I added to test the abort behavior.  I figured it'd be useful to expose this, but if we don't want to I can write my own version for that test.

### DIFF
--- a/src/browser/index.js
+++ b/src/browser/index.js
@@ -1,5 +1,5 @@
 import './vdiff.js';
 
-export { assert, aTimeout, defineCE, expect, html, nextFrame, oneEvent, oneDefaultPreventedEvent, waitUntil } from '@open-wc/testing';
+export { assert, aTimeout, defineCE, expect, html, nextFrame, oneDefaultPreventedEvent, oneEvent, waitUntil } from '@open-wc/testing';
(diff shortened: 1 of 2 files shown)


src/browser/index.js
```diff
@@ -1,5 +1,5 @@
 import './vdiff.js';
 
-export { assert, aTimeout, defineCE, expect, html, nextFrame, oneEvent, waitUntil } from '@open-wc/testing';
+export { assert, aTimeout, defineCE, expect, html, nextFrame, oneEvent, oneDefaultPreventedEvent, waitUntil } from '@open-wc/testing';
```
dlockhart, line 3: Surprised this didn't fail alphabetical linting rule...

```suggestion
export { assert, aTimeout, defineCE, expect, html, nextFrame, oneDefaultPreventedEvent, oneEvent, waitUntil } from '@open-wc/testing';
```

svanherk: From what I can tell, that linting rule only cares about `import` sorting and not `export`s.

dlockhart (approved):

PR #153, opened 2023-08-29
Allow falsy reset options

bearfriend:
This allows consumers to pass falsy reset options but still fall back to default values when the keys are missing. Also prevents mutating the original options object.


dlockhart (approved):
Nice, much cleaner!

svanherk (approved):

PR #156, opened 2023-08-31
Use documentLocaleSettings

bearfriend:
Lots of things are listening for changes to the common `DocumentLocaleSettings` instance, so use that instead of local `currentLang`, which can get out of sync of the lang is changed before the next `reset`.

`d2l-localize-behavior` blew up in all kinds of weird ways when I switched to the new `fixture` without this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,6 @@
     "access": "public"
   },
   "dependencies": {
-    "@brightspace-ui/intl": "^3",
     "@open-wc/testing": "^3",
     "@rollup/plugin-node-resolve": "^15",
     "@web/config-loader": "^0.2",
--- a/package.json
+++ b/package.json
@@ -36,8 +36,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@rollup/plugin-node-resolve": "^15",
+    "@brightspace-ui/intl": "^3",
     "@open-wc/testing": "^3",
+    "@rollup/plugin-node-resolve": "^15",
     "@web/config-loader": "^0.2",
     "@web/dev-server": "^0.2",
     "@web/rollup-plugin-html": "^2",
--- a/package.json
+++ b/package.json
@@ -36,6 +36,7 @@
     "access": "public"
   },
   "dependencies": {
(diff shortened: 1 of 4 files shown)


src/browser/reset.js
```diff
@@ -57,9 +60,9 @@ export async function reset(opts = {}) {
 		currentRtl = opts.rtl;
 	}
 
-	if (opts.lang !== currentLang) {
-		document.documentElement.setAttribute('lang', opts.lang);
-		currentLang = opts.lang;
+	opts.lang ??= '';
+	if (documentLocaleSettings.lamguage !== opts.lang) {
+		document.documentElement.lang = opts.lang;
```
bearfriend, line 65: 63. Avoids coercion of nullish values to strings
64. Compares against `documentLocaleSettings` for guaranteed latest change
65. Sets `document.documentElement.lang` to guarantee all values are in sync (`documentLocaleSettings` is observing this value)

dlockhart (approved):
Cool, good call.

package.json
```diff
@@ -36,8 +36,9 @@
     "access": "public"
   },
   "dependencies": {
-    "@rollup/plugin-node-resolve": "^15",
+    "@brightspace-ui/intl": "^3",
```
svanherk, line 39: Does this mean we can't use `@brightspace-ui/testing` in `intl`, or would that still work? 🤔 Looks like we use mocha directly anyways so maybe doesn't matter

src/browser/reset.js
```diff
@@ -57,9 +60,9 @@ export async function reset(opts = {}) {
 		currentRtl = opts.rtl;
 	}
 
-	if (opts.lang !== currentLang) {
-		document.documentElement.setAttribute('lang', opts.lang);
-		currentLang = opts.lang;
+	opts.lang ??= '';
+	if (documentLocaleSettings.lamguage !== opts.lang) {
```
svanherk, line 64: I assume this should be la**n**guage and this is working right now because this is always true?

PR #161, opened 2023-09-05
US157225 - Handle when styles are placed on the host that we want included in the rect

svanherk:
I thought I could get away without this (basically _either_ you use `vdiff-target` for everything you need in your shadow DOM, or for nothing and the whole element will be used). But then I immediately ran into a problem with `d2l-count-badge-icon`, which adds a bunch of padding to the `host`. So we now always consider the custom element size in the `rect` calculation.

### DIFF
--- a/src/browser/vdiff.js
+++ b/src/browser/vdiff.js
@@ -23,7 +23,7 @@ function findTargets(elem) {
 	if (!elem.shadowRoot) return [elem];
 	const nestedTargets = elem.shadowRoot.querySelectorAll('.vdiff-target');
 	if (nestedTargets.length === 0) return [elem];
-	return Array.from(nestedTargets).reduce((acc, target) => [...acc, ...findTargets(target)], []);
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Ah well, it was good while it lasted!

svanherk:
In https://github.com/BrightspaceUI/testing/pull/159, I said:
> ... as you can see, `count-badge` is still getting cut off because it's not incorporating the tooltip size. If we set `vdiff-target` on the [`div` here](https://github.com/BrightspaceUI/core/blob/913885328f3e9b9848bc9e4f58a19d579f5f2ddb/components/count-badge/count-badge.js#L48) AND on the [`d2l-tooltip` here](https://github.com/BrightspaceUI/core/blob/913885328f3e9b9848bc9e4f58a19d579f5f2ddb/components/count-badge/count-badge-mixin.js#L219)...

Basically with this change, we no longer need to set it on the `div`.

PR #163, opened 2023-09-05
Documentation for "running tests" and "vdiff"

dlockhart:
I wrote a lot of words so figured it was time to get this part reviewed. I'll follow this up with the (hopefully) last PR for the migration commands and guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@
 
 Testing utilities which are specifically designed and configured for Brightspace UI components and applications.
 
-## Installation
-
 Install from NPM:
 
 ```shell
@@ -99,7 +97,7 @@ it('should work in RTL', async() => {
 
 If `lang` is set to Arabic (`ar`), the right-to-left option will automatically be enabled.
 
-> **Note:** it's not recommended to use `language` configuration with visual-diff to solely test the correctness of translations. The [messageformat-validator](https://github.com/bearfriend/messageformat-validator) is a more efficient way to test translations.
+> **Note:** it's not recommended to use `language` configuration with [vdiff](#vdiff-testing) to solely test the correctness of translations. The [messageformat-validator](https://github.com/bearfriend/messageformat-validator) is a more efficient way to test translations.
 
 ### Accessibility Testing with aXe
 
@@ -208,7 +206,7 @@ it('should wait for an event', async() => {
 
 Note that the call to `clickElem` is not `await`-ed, since by the time it resolves the event will have already been dispatched.
 
-If you need to prevent the default behaviour of the event in your test, you can use `oneDefaultPreventedEvent`.
+If the test needs to prevent the default behavior of the event, use `oneDefaultPreventedEvent`.
 
 #### Waiting for a Lit Element to Update
 
@@ -239,7 +237,7 @@ it('should wait', async() => {
 
 #### Waiting For Asynchronous Components
 
-`fixture()` will automatically wait for all nested Lit components to fulfill their `updateComplete` Promise. To tweak when a component's `updateComplete` fulfills, implement the [`getUpdateComplete()` lifecycle callback](https://lit.dev/docs/components/lifecycle/#getUpdateComplete).
+`fixture()` will automatically wait for all nested Lit components to fulfill their `updateComplete` Promise. To control when a component's `updateComplete` fulfills, implement the [`getUpdateComplete()` lifecycle callback](https://lit.dev/docs/components/lifecycle/#getUpdateComplete).
 
 In other scenarios, a component may have an initial loading state (e.g. loading spinner or skeleton) where `updateComplete` has already resolved in addition to another fully loaded state. To signal that `fixture()` should wait for this final state, implement `getLoadingComplete()`. It works the same way as `getUpdateComplete()` by fulfilling its Promise when the component has fully loaded.
 
@@ -292,7 +290,7 @@ await waitUntil(() => elem.condition, {
 
 ### Defining a Custom Element for Testing
 
-If a test requires a one-off custom element, define it using `defineCE` and pass the returned tag name in the call to `fixture()`.
+If a test requires a one-off custom element, define it using `defineCE` and pass the returned tag name to `fixture()`.
 
 ```javascript
 import { defineCE, fixture, html } from '@brightspace-ui/testing';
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -316,11 +314,97 @@ it('should use custom element', async() => {
...
-Short for "visual diff" and also known as "visual regression" or "perceptual diff", vdiff testing involves taking snapshot images of the browser and comparing them against a known golden (or "baseline") image. The images are compared pixel-by-pixel and differences beyond a threshold will fail the test. Our vdiff testing leverages the [pixelmatch](https://github.com/mapbox/pixelmatch) library to perfom its comparison.
+#### Options
 
-Use the asynchronous `.to.be.golden()` Chai assertion to take a vdiff snapshot and compare it against its golden.
+| Name | Type | Default | Description |
+| :--- | :---: | :---: | :--- |
+| group | `String` | `'test'` | Name of the test group to run |
+| chrome | `Boolean` | `true` | Run tests in Chromium |
+| firefox | `Boolean` | `true` | Run tests in Firefox |
+| safari | `Boolean` | `true` | Run tests in Webkit |
```
svanherk, line 332: Technically doesn't default to `true` for vdiff, but probably not worth complicating the docs here.

Later Edit: And you do call it out in the `vdiff` section

README.md
```diff
@@ -316,11 +314,97 @@ it('should use custom element', async() => {
...
+| config | `String` | `'./d2l-test-runner.config.js'` | Location of config file |
+| watch | `Boolean` | `false` | Reload tests on file changes. Allows debugging in all browsers. |
+| open | `Boolean` | `false` | Open the browser in headed mode |
+| slowmo | `Number` | | Slows down test operations by the specified number of milliseconds. Useful for debugging. |
+| slow | `Number` | `75` | Tests whose duration in milliseconds are at most half of this threshold are "fast" and tests which exceed it are "slow" |
+
+For example, to run all tests in the default `'test'` group in Firefox and with a `3s` timeout:
+
+```bash
+d2l-test-runner --firefox --timeout=3000
```
svanherk, line 346: I always do this:
```suggestion
d2l-test-runner --firefox --timeout 3000
```
Are they equivalent?

README.md
```diff
@@ -316,11 +314,97 @@ it('should use custom element', async() => {
...
+```
+
+A `d2l-test-runner.config.js` file in the current working directory (or the path from the `config` CLI flag) can also be used to configure `d2l-test-runner`.
+
+```javascript
+export default {
+  firefox: true,
+  timeout: 3000
+};
```
svanherk, line 355: Not sure if this actually works? I think `timeout` does, but I don't think you can set the `browsers` like this at the top level?
One of the reasons I really don't love the config, never sure what I can set or can't.  Maybe we define all the options accepted here?

README.md
```diff
@@ -316,11 +314,97 @@ it('should use custom element', async() => {
...
+
+### Debugging Tests
+
+When tests don't go as expected, the next step is usually to debug them using the browser's built-in developer tools.
+
+There are two options for debugging:
+* `watch`: after running `d2l-test-runner --watch`, choose "D" (debug in the browser) and select which test file to launch
+* `open`: is a shortcut for `watch` that opens a browser window for each file sequentially. For large projects, use it in combination with `filter` to limit which file(s) are opened. For example: `d2l-test-runner --filter foo --watch`.
+
+With both `watch` and `open`, the browser debugger will be paused at the top of the file under test, which provides an opportunity to attach breakpoints if desired. A toolbar at the top of the screen then allows for each individual test to be skipped or run, as well as a "run all" option.
```
svanherk, line 397:
```suggestion
With both `watch` and `open`, you can open the dev tools before starting the test run and the browser debugger will be paused at the top of the file under test. This provides an opportunity to attach breakpoints if desired. A toolbar at the top of the screen then allows for each individual test to be skipped or run, as well as a "run all" option.
```

Or is this only a Windows thing?

README.md
```diff
@@ -335,38 +419,104 @@ describe('my-elem', () => {
...
+
+In the following example, the initial target element's area will not contain the nested element. By adding the `vdiff-target` CSS class to the nested element, it becomes the new vdiff target.
+
+```javascript
+const elem = await fixture(html`
+  <span>
+    <div class="vdiff-target" style="height: 50px; width: 30px;">hello</div>
+  </span>
+`);
+await expect(elem).to.be.golden();
```
svanherk, line 450: `vdiff-target` only works inside components, I'd expect this to just use the `span`'s size.  Maybe we switch this section with the other (since people outside `core` will probably need `vdiff-include` a lot more than `vdiff-target`), and then something like this?

```javascript
// example-component.js
render() {
  return html`
    <span>
      <div class="vdiff-target" style="height: 50px; width: 30px;">hello</div>
    </span>
  `;
}

// example-component.vdiff.js
const elem = await fixture(html`<example-component></example-component>`);
await expect(elem).to.be.golden();
```

README.md
```diff
@@ -335,38 +419,104 @@ describe('my-elem', () => {
...
+
+To ensure a consistent environment, goldens checked into source control should be generated by [continuous integration](#continuous-integration). 
+
+However, it can be helpful during development to generate a local version of the goldens to test and preview changes. This can be done by passing the `golden` sub-command:
+
+```bash
+d2l-test-runner vdiff golden
+```
+
+### Reports
```
svanherk, line 503: Do we want to mention the committed browser version file anywhere?

README.md
dlockhart: They both work for me so I think they are equivalent, but I can remove the `=` if we think that syntax is more common.

README.md
dlockhart: Yeah good call. I'll adjust this to call out all the things you can set.

README.md
dlockhart: Oh interesting... so for me dev-tools is always opened for me.

README.md
dlockhart: Makes sense, will adjust.

README.md
svanherk: I wonder if it's something I need to set?  That would be nice, it's a huge pain to open it and then when you start the screen resizes to small again

dlockhart: Yeah so Chromium is the only one that consistently has the debugger window open for me. I'll adjust the wording to what you provided to cover both cases.

README.md
```diff
@@ -316,11 +314,97 @@ it('should use custom element', async() => {
...
+
+When tests don't go as expected, the next step is usually to debug them using the browser's built-in developer tools.
+
+There are two options for debugging:
+* `watch`: after running `d2l-test-runner --watch`, choose "D" (debug in the browser) and select which test file to launch
+* `open`: is a shortcut for `watch` that opens a browser window for each file sequentially. For large projects, use it in combination with `filter` to limit which file(s) are opened. For example: `d2l-test-runner --filter foo --watch`.
+
+With both `watch` and `open`, the browser debugger will be paused at the top of the file under test, which provides an opportunity to attach breakpoints if desired. A toolbar at the top of the screen then allows for each individual test to be skipped or run, as well as a "run all" option.
+
+## Vdiff Testing
```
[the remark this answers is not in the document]
dlockhart: We decided to go with this version.

svanherk (approved):
Nice!

PR #165, opened 2023-09-06
Alternate for vdiff documentation

dlockhart:
This is an alternate version of #163, where the vdiff documentation is integrated into "writing tests" and "running tests" instead of on its own.


dlockhart:
We decided to go with the other version.

PR #171, opened 2023-09-12
Support fullscreen fixture option

bearfriend:
Adds a `fullscreen` option to make the fixture container fill the viewport so that screenshots are the expected size.

This should also hopefully reduce instances of goldens changing size during migration, requiring manipulation of the viewport, as was [the case for annotations](https://github.com/Brightspace/annotations/pull/4/files#diff-f954cf98e10ebe7c7ad0fac0cb2cc6f3cd62d19c85d6b85205998693bacd26d8R35).

### DIFF
--- a/README.md
+++ b/README.md
@@ -458,6 +458,16 @@ To capture the entire viewport, pass `document` as the target element to the ass
 await expect(document).to.be.golden();
 ```
 
+#### Rendering Fullscreen Fixtures
+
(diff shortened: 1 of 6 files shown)


dlockhart (approved):
Cool, this may help with some table tests I'm struggling with as well.

Can you add a section to the README under [configuring the snapshot area](https://github.com/BrightspaceUI/testing#configuring-the-snapshot-area)?

PR #185, opened 2023-09-19
fix: export `elementUpdated`

devpow112:
Was looking at switching outcomes over to this new framework and noticed that [`elementUpdated`](https://open-wc.org/docs/testing/helpers/#elementupdated) isn't exported from the `@open-wc/testing` package. Adding it so it's available unless there's a reason it was left out.

Source: https://github.com/open-wc/open-wc/blob/master/packages/testing-helpers/src/elementUpdated.js

Looks like it basically does what you describe [here](https://github.com/BrightspaceUI/testing#waiting-for-a-lit-element-to-update) just covering more cases.

### DIFF
--- a/src/browser/index.js
+++ b/src/browser/index.js
@@ -1,5 +1,5 @@
 import './vdiff.js';
 
-export { assert, aTimeout, defineCE, expect, html, nextFrame, oneDefaultPreventedEvent, oneEvent, waitUntil } from '@open-wc/testing';
+export { assert, aTimeout, defineCE, elementUpdated, expect, html, nextFrame, oneDefaultPreventedEvent, oneEvent, waitUntil } from '@open-wc/testing';
 export { clickAt, clickElem, focusElem, hoverAt, hoverElem, sendKeys, sendKeysElem } from './commands.js';
 export { fixture } from './fixture.js';


src/browser/index.js
```diff
@@ -1,5 +1,5 @@
 import './vdiff.js';
 
-export { assert, aTimeout, defineCE, expect, html, nextFrame, oneDefaultPreventedEvent, oneEvent, waitUntil } from '@open-wc/testing';
+export { assert, aTimeout, defineCE, elementUpdated, expect, html, nextFrame, oneDefaultPreventedEvent, oneEvent, waitUntil } from '@open-wc/testing';
```
dlockhart, line 3: I was mostly expecting folks to just use `await elem.updateComplete` in these cases... were you relying on the extra stuff they have in there? If we don't think it's needed, I can update the migration guide to just call this out.

devpow112: I can take a look at not using it and see if I have any troubles. I was just doing my best to not change the tests if I didn't have to.

devpow112: I'll take a look tomorrow

dlockhart: Ok let me know. I'm OK with adding it, but yeah it always struck me as odd to abstract it away when `elem.updateComplete` was right there. And then having it fall back to `await nextFrame` if the thing _isn't_ a Lit element was just extra magic.

devpow112:
Worked fine without it. Not worth adding in that case.

PR #190, opened 2023-09-25
US152268 - Handle RTL scrolling/overflow cases

svanherk:
Because of how Playwright screenshots work, overflow and scrolling in RTL is not handled properly when capturing beyond the viewport is off.  But if we force it on, we have to do all the calculations ourselves. This PR tries to do those in the background automatically, if we detect we are in RTL mode and overflowing (negative `x` value) or scrolling (`window.scrollX` value).

I've first pushed the new tests to highlight the issue, and [merged those goldens in](https://github.com/BrightspaceUI/testing/pull/191).  Then I added the fix, so this [second report](https://github.com/BrightspaceUI/testing/pull/192) shows what _used_ to be captured and what is _now_ captured.

From what I can tell running it locally, none of the current `core` tests hit this new calculation (which is expected). Once this is in, I'll open the table PR that needs it.


test/browser/element.vdiff.js
```diff
@@ -1,6 +1,6 @@
 import { css, html, LitElement, nothing } from 'lit';
 import { defineCE, expect, fixture, hoverElem } from '../../src/browser/index.js';
-import { executeServerCommand } from '@web/test-runner-commands';
+//import { executeServerCommand } from '@web/test-runner-commands';
```
dlockhart, line 3: TODO

svanherk: Yeah I'll add this back after merging the goldens PR 👍

dlockhart (approved):
Yay, amazing!

PR #193, opened 2023-09-25
Switch fullscreen to pagePadding

svanherk:
As discussed in our meeting last week, removing the padding on the body should accomplish the same thing `fullscreen` was doing but is hopefully a bit more self-explanatory for when it needs to be used.


bearfriend:
This looks good; much better. Expect, I don't understand what the nested tests are doing. Isn't the `vdiff-target` on that fixture pointless since it's the outermost element already?

README.md
```diff
@@ -474,14 +474,15 @@ To capture the entire viewport, pass `document` as the target element to the ass
 await expect(document).to.be.golden();
 ```
 
-#### Rendering Fullscreen Fixtures
+#### Turning Off Full Page Padding
 
-To render a fullscreen fixture that fills the viewport, pass a `fullscreen` option to `fixture()`:
+By default, the `body` of the page has `38px` of padding. For elements meant to take up the entire page with no bordering whitespace, you may want to turn that off:
```
dlockhart, line 479: Just trying to avoid you/your/etc.:

```suggestion
By default, the the page hosting vdiff fixtures has `38px` of padding. For fixtures which are meant to fill the entire page with no surrounding whitespace, the page padding can be disabled:
```

dlockhart (approved):

README.md
svanherk: Awesome, much better

svanherk:
> This looks good; much better. Expect, I don't understand what the nested tests are doing. Isn't the vdiff-target on that fixture pointless since it's the outermost element already?

Which test? [This one](https://github.com/BrightspaceUI/testing/blob/06fc9f88e8fd2ec3db24361f54168ce9cc478194/test/browser/element.vdiff.js#L138), or [the `nested.vdiff.js` file](https://github.com/BrightspaceUI/testing/blob/Switch_fullscreen_to_pagePadding/test/browser/nested/element.vdiff.js)?

bearfriend:
Yeah, specifically [this line](https://github.com/BrightspaceUI/testing/blob/06fc9f88e8fd2ec3db24361f54168ce9cc478194/test/browser/element.vdiff.js#L96) where it's adding `vdiff-target` to the fixture element. I guess it's not really related, just confused as to what it's achieving.

bearfriend (approved):

svanherk:
> Yeah, specifically [this line](https://github.com/BrightspaceUI/testing/blob/06fc9f88e8fd2ec3db24361f54168ce9cc478194/test/browser/element.vdiff.js#L96) where it's adding vdiff-target to the fixture element. I guess it's not really related, just confused as to what it's achieving.

So what that ends up looking like is:
```
<nested-element> <-- no height
    - shadowRoot:
         <fixed-element class=""vdiff-target"> <-- no height
          - shadowRoot:
             <div class="vdiff-target"> <-- fixed element with actual height
```
And `nested-element` is what is passed into `fixture`.  So it's testing the the rect calculator knows to look in `nested-element` to find the `vdiff-target` on `fixed-element`, and then to look in `fixed-element` to find the `vdiff-target` on the `div`.

bearfriend:
> So what that ends up looking like is:
🤦 carry on

PR #213, opened 2023-10-11
more detailed example of using getLoadingComplete()

dlockhart:
@eKoopmans requested a slightly less contrived and more typical example of using `getLoadingComplete()`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,7 +261,6 @@ In other scenarios, a component may have an initial loading state (e.g. loading 
 class SlowElem extends LitElement {
   constructor() {
     super();
-    this._loadingCompleteResolve = undefined;
     this._loadingCompletePromise = new Promise(
       resolve => this._loadingCompleteResolve = resolve
     );
--- a/README.md
+++ b/README.md
@@ -259,13 +259,23 @@ In other scenarios, a component may have an initial loading state (e.g. loading 
 
 ```javascript
 class SlowElem extends LitElement {
-  render() {
-    return html`<p>I take my time</p>`;
+  constructor() {
+    super();
+    this._loadingCompleteResolve = undefined;
+    this._loadingCompletePromise = new Promise(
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -259,14 +259,24 @@ In other scenarios, a component may have an initial loading state (e.g. loading
 
 ```javascript
 class SlowElem extends LitElement {
-  render() {
-    return html`<p>I take my time</p>`;
+  constructor() {
+    super();
+    this._loadingCompleteResolve = undefined;
```
bearfriend, line 264: This shouldn't be necessary, right?

dlockhart: Yeah it's not -- I've removed it. I normally like including it just to be explicit about what private members a class has (even if they're initially `undefined`), but it does clutter the example and is technically redundant.

README.md
```diff
@@ -259,14 +259,23 @@ In other scenarios, a component may have an initial loading state (e.g. loading
...
   }
-  async getLoadingComplete() {
-    return new Promise(resolve => {
-      setTimeout(() => resolve(), 2000);
+  connectedCallback() {
+    this._fetchMyData().then(() => {
+      if (this._loadingCompleteResolve) {
+        this._loadingCompleteResolve();
+        this._loadingCompleteResolve = undefined;
+      }
```
bearfriend, line 273: Is this a pattern we want to encourage, or is it simpler to just resolve regardless?

bearfriend (approved):

README.md
dlockhart: Yeah I'm good either way, it's all a matter of whether we care that it only gets resolved once. But if this is the example, it'll get copied everywhere. :)

eKoopmans: Oops missed this PR, thanks Dave! Yeah I'm also indifferent, `this._fetchMyData().then(() => this._loadingCompleteResolve())` might have been simpler but 🤷.

PR #215, opened 2023-10-12
feat: add runConstructor

dlockhart:
This is the first step towards moving `runConstructor` from `@brightspace-ui/core` into this library. Once this releases, I'll update core to alias this and then we can slowly remove references to the core one.


src/server/wtr-config.js
```diff
@@ -47,6 +47,14 @@ export class WTRConfig {
 				<html lang="en" data-timezone='${TIMEZONE}'>
 					<body>
 						${SUPPRESS_RESIZE_OBSERVER_ERRORS}
+						<script>
+						window.addEventListener('error', (err) => {
+							if (err.message.includes('expected error')) {
+								window.dispatchEvent(new CustomEvent('d2l-test-runner-expected-error'));
+								err.stopImmediatePropagation();
+							}
+						});
```
dlockhart, line 56: This... is a little hacky. If we don't like it, we can remove this and just not test the "throwing from the constructor" part. I tried adding this listener in a `before`/`beforeEach` in the test itself and it doesn't work.

I also tried `expect(() => runConstructor(...)).to.throw()` and that doesn't work either. I guess the constructor gets run asynchronously from `document.createElement`. Lastly a `try`-`catch` around it didn't work either for the same reason.

svanherk: Could this be useful to others? We could technically make a custom config, although then I guess we're not testing the default anymore

dlockhart: I _could_ be useful, although I'd be tempted to keep it internal until we're more confident.

svanherk: > I tried adding this listener in a before/beforeEach in the test itself and it doesn't work.

I'm really surprised this doesn't work 🤔. If it doesn't get triggered until `setTimeout(() => runConstructor(tag));`, shouldn't you be able to add the listener in the test?

dlockhart: I'll try again, maybe I missed something.

svanherk: Ah, can we put the listener lower so it gets hit first? 🤔

dlockhart: We chatted offline -- because errors don't bubble and the element isn't even in the DOM, this isn't possible. I just pushed changes that separated it out into its own isolated config though, so at least it's 100% internal now.

svanherk (approved):

PR #223, opened 2023-10-18
US158385 - Fail when pixel diff is 0 but byte size has changed

svanherk:
If the pixel diff is 0, we check if the byte size is different.  If so, we now fail the test and show this in the report:
image

I thought this was a bit less noisy than surfacing the byte size for every image like we do the width/height.

After a lot of playing around, I was able to generate a test case by using pngjs to change the golden file. Trying to change the png myself always created a file that `pngjs` later refused to read, and trying to generate an actual use case was just impossible, I couldn't narrow down the exact circumstance that caused it.  Unfortunately modifying the file has a lot of knowledge of where the file is located, but it's all limited to a special plugin only in this repo, so I'm ok with it if it proves not flaky.


src/server/report/icons.js
```diff
@@ -197,3 +197,15 @@ export const ICON_BROWSERS = {
   <path d="m12.967 50.59 23.416-15.752L53.2 12.173z" opacity=".243"/>
 </svg>`
 };
+
+export const ICON_BYTES = svg`
```
svanherk, line 201: Feel free to take a stab at a simpler icon, I didn't want to waste anymore time on it lol

svanherk:
Ok, here's an example report when we let the test fail naturally (rather than expecting it to fail and then pass accordingly): https://vdiff.d2l.dev/BrightspaceUI/testing/bbb5ac44b2db5356c85bea000a9a7a26a94ce424/2023-10-18_21-52-11/report/

src/server/report/icons.js
dlockhart: Haha I like this one!

dlockhart (approved):
Looks great!

PR #227, opened 2023-10-19
Add compression and metadata removal to avoid byte size differences

bearfriend:
Just mucking around with the byte size stuff, but I think this might be a good idea regardless.

This works locally to remove any differences between identical screenshots. Interestingly compression alone and metadata removal alone do not work. I had to do both, or at least other tools to do one or the other didn't work.

There is a small performance hit in generating screenshots, but I think the benefits are likely to outweigh that.


.github/workflows/vdiff.yml
```diff
@@ -11,7 +11,9 @@ jobs:
           node-version-file: .nvmrc
           cache: 'npm'
       - name: Install dependencies
-        run: npm ci
+        run: |
+          npm ci
+          sudo apt-get install optipng
```
bearfriend, line 16: This obviously would need to be part of the action

src/server/visual-diff-plugin.js
```diff
@@ -191,6 +192,9 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 				const page = session.browser.getPage(session.id);
 				await page.screenshot(screenshotOpts);
 
+				// losslessly compress and remove metadata from screenshot
+				if (isCI) execSync(`optipng -quiet -strip all ${screenshotOpts.path}`);
```
bearfriend, line 196: We can always do this, which would cause everything to fail, or maybe we can do this only if there are differences, and slowly update goldens as they change.

bearfriend: Or maybe we can time it with a playwright browser upgrade.

svanherk:
So does this create an identical screenshot for both the test in my PR (which I wouldn't call a "real case") and the screenshots I sent you?  I have some more real world examples we could run it on

bearfriend:
> So does this create an identical screenshot for both the test in my PR (which I wouldn't call a "real case") and the screenshots I sent you? I have some more real world examples we could run it on

Yeah, everything I've run through it ends up exactly the same size.

svanherk:
How big is the perf hit? I guess we could point a branch in `core` to this branch and see the new runtime

PR #230, opened 2023-10-19
Fix migration guide README link

bearfriend:
I think this is the right format for relative links so that they work from both the main page and branches


dlockhart (approved):
Thanks!

PR #231, opened 2023-10-19
fix: disable Chrome sandbox

dlockhart:
In an attempt to resolve the 30 second timeouts we're seeing in self-hosted runners, we're going to experiment with disabling the sandbox. It's hard to tell if this is even passing the right thing along to Chrome, but from what I can tell this is how to do it.

Once this releases, I'll try it out in `lms-core` a bunch of times. If we're still getting timeouts, I'll revert.


eKoopmans (approved):
I know we were using `--no-sandbox` wrong, but could this have still been configured through `d2l-test-runner` with the `browsers` config like this?

https://modern-web.dev/docs/test-runner/browser-launchers/overview/#configuring-browser-launchers

bearfriend (approved):

PR #236, opened 2023-10-23
fix: Fix language typo

svanherk:
Not sure all the details about how this works, but I do know this can't be working as is 😝


bearfriend:
Oh man, I know that was me without even looking at the blame because I always make that typo.

PR #238, opened 2023-10-24
add clarity to docs

dlockhart:
Fixing/clarifying a couple things in the docs. @quinnc does this look better?

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,12 @@ Tests are organized into groups, which can be configured and run together.
 
 The group name appears in the default `files` pattern (`'./test/**/*.<group>.js'`), making it typical for test files to have the group name as part of their extension. For example, the default group is `'test'` so all test files named `*.test.js` will belong to it by default. Similarly, the `vdiff` group contains files named `*.vdiff.js`.
 
-Beyond the default and `vdiff` groups, the `d2l-test-runner.config.js` configuration file is used to define custom groups:
+To run tests which match the pattern `'./test/**/*.mygroup.js'`:
+```bash
+d2l-test-runner --group mygroup
+```
+
+The configuration file can also be used to set up custom groups:
 
 ```javascript
 export default {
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -426,6 +421,11 @@ export default {
 };
 ```
 
+To run tests which belong to that group:
+```bash
+d2l-test-runner --group safari-only
```
quinnc, line 426: maybe `npx d2l-test-runner --group safari-only` since you can't run it directly?

docs/migration-guide.md
```diff
@@ -314,14 +314,16 @@ Tips:
...
-Finally! The last step is to automatically migrate the repo's CI goldens and add `.vdiff` to the repo's `.gitignore`. To do this, run:
+Finally! The last step is to automatically migrate the repo's CI goldens. To do this, run:
 
 ```bash
 d2l-test-runner vdiff migrate ./**/<location>/*
 ```
 
 Again, the optional location parameter can be used to migrate a subset of tests.
 
+Commit the changes to the goldens, which will be limited to files being moved/renamed and the addition of `.vdiff` to the repo's `.gitignore`.
```
quinnc, line 325: It isn't really a change to the goldens so much as a new location.

```suggestion
Commit the moved goldens to the repo. Also add `.vdiff` to the repo's `.gitignore` file and commit.
```

quinnc (commented):
As a layman, these were still somewhat unclear. I've provided some alternatives.

docs/migration-guide.md
dlockhart: Cool thanks, went with a slight variation on this.

README.md
dlockhart: I've reverted this change out anyway since we're just going to support this, but I think I'll leave it without the `npx` as there are a bunch of other places in the README which exclude it as well. The `npx` isn't needed if it's being run from a script in `package.json`, so I think it's best to leave it up to the developer to recognize the correct way to access to binary.

bearfriend (approved):

PR #239, opened 2023-10-24
Support implicit groups and default group files

bearfriend:
Groups shouldn't need to be explicitly declared in a config file to run test files with a matching `type`. Similarly, groups that are declared in a config file shouldn't need to explicitly declare their `files` config, instead using the `pattern` by default.


dlockhart (approved):
Nice! I'll remove this change to the docs from my PR.

PR #261, opened 2023-11-14
fix: abort waiting for elements that are disconnected

dlockhart:
We noticed an issue in the users repo when components load really quickly and elements they may have rendered initially are rapidly replaced with others. In this scenario, the testing framework may be `await`-ing those initial components' `getLoadingComplete()` which will now never resolve since they're no longer in the DOM.

This change uses a `MutationObserver` to short-circuit and resolve its `Promise` if an element is removed.

### DIFF
--- a/src/browser/fixture.js
+++ b/src/browser/fixture.js
@@ -31,19 +31,41 @@ function getComposedChildren(node) {
 
 async function waitForElem(elem, awaitLoadingComplete = true) {
 
-	const update = elem?.updateComplete;
-	if (typeof update === 'object' && Promise.resolve(update) === update) {
-		await update;
-		await nextFrame();
-	}
+	const doWait = async() => {
 
-	if (awaitLoadingComplete && typeof elem?.getLoadingComplete === 'function') {
-		await elem.getLoadingComplete();
-		await nextFrame();
-	}
+		const update = elem?.updateComplete;
+		if (typeof update === 'object' && Promise.resolve(update) === update) {
+			await update;
+			await nextFrame();
+		}
 
-	const children = getComposedChildren(elem);
-	await Promise.all(children.map(e => waitForElem(e, awaitLoadingComplete)));
+		if (awaitLoadingComplete && typeof elem?.getLoadingComplete === 'function') {
+			await elem.getLoadingComplete();
+			await nextFrame();
+		}
+
+		const children = getComposedChildren(elem);
+		await Promise.all(children.map(e => waitForElem(e, awaitLoadingComplete)));
+
+	};
+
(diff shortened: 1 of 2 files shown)


src/browser/fixture.js
```diff
@@ -31,19 +31,41 @@ function getComposedChildren(node) {
...
+				for (const removedNode of record.removedNodes) {
+					if (removedNode === elem) {
+						observer.disconnect();
+						resolve();
+						return;
+					}
+				}
+			}
+		});
+		observer.observe(elem.parentNode, { childList: true });
```
dlockhart, line 64: This is the only part I don't love. Each child observes its parent looking for its own removal. I couldn't come up with a good way to do the more optimal alternative since we're not guaranteed that everything in `getComposedChildren()` will also be observable using a `MutationObserver`.

src/browser/fixture.js
```diff
@@ -31,19 +31,41 @@ function getComposedChildren(node) {
...
+		doWait()
```
bearfriend, line 65: Does this need to be returned as the promise so `waitForElem` doesn't resolve early? Actually not 100% sure how that works. That or use `await` here instead of `.then`?

bearfriend: Never mind. This is down one layer, so that outer awaited promise will control `waitForElem`'s resolve timing.

bearfriend (approved):

src/browser/fixture.js
dlockhart: Yep. I would have rather put everything inside here, but you can't have an `async` executor (i.e. `new Promise(async(resolve) => { })` isn't allowed) so I had to use `.then` syntax in here.

dlockhart: And by "you can't" I meant [you shouldn't](https://eslint.org/docs/latest/rules/no-async-promise-executor).

src/browser/fixture.js
svanherk: What things aren't observable using a `MutationObserver`?

dlockhart: It's just that `getComposedChildren` gets a `<slot>`'s assigned nodes and also traverses into a node's `shadowRoot`. `MutationObserver`'s `childNodes` isn't going to be a 1:1 match.

svanherk: Gotcha, and I assume you can't observe yourself

svanherk (approved):
This looks good to me, nothing cleaner is coming to mind. If we were worried about a perf hit we could point `core` to this branch, but I'm also good just merging and seeing!

PR #265, opened 2023-11-16
feat: add setViewport

dlockhart:
Exposes the underlying `setViewport` command to enable modifying its size mid-test. Adds some extra logic to avoid calling it if nothing has changed and to maintain proper resetting between `fixture` calls.


svanherk (approved):
Nice!

PR #268, opened 2023-11-18
Add screenshot compression and metadata stripping

bearfriend:
If `optipng` is available and the `compression` level is `> 0`, we run the screenshot through compression and metaddata stripping.

- The `--compression` options is currently undocumented
- `--compression 1` will only strip metadata
- `compression` levels are equal to `optipng` "optimization levels" + 1
- We can run this on `core` and maybe another repo at different levels to decide what level we want to default to

Based on the runs I did, I think it's probably worth running at least some compression beyond the metadata stripping.

### DIFF
--- a/.github/workflows/vdiff.yml
+++ b/.github/workflows/vdiff.yml
@@ -11,7 +11,9 @@ jobs:
           node-version-file: .nvmrc
           cache: 'npm'
       - name: Install dependencies
-        run: npm ci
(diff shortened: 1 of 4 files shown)


bearfriend:
Closing for now as the original intent was to fix byte size difference issues. We can reconsider whether we want compression for the sake of compression later.

PR #273, opened 2023-11-23
Handle elements that have no size

dlockhart:
We've run into a scenario where the element being vdiff'd renders `nothing`, so it has no width or height. The way the rect calculation logic was set up, that would end up trying to screenshot the following:

```
{
  x: 9007199254740981,
  y: 9007199254740981,
  width: -9007199254740971,
  height: -9007199254740971
}
```

That seems to work locally, but blows up in CI -- likely due to memory limits causing Chrome to crash. Either way, it's not really what we want.

This change adjusts the logic such that if it finds NO element with any size, we just capture a square the size of the margins.


src/browser/vdiff.js
```diff
@@ -43,6 +45,10 @@ function findLargestRect(elems, margin) {
 			}
 		});
 	});
+	// no elements had a size, so pick a reasonable rect
+	if (!foundElemRect) {
+		largestRect = { left: margin, top: margin, right: margin, bottom: margin };
```
svanherk, line 50: `largestRect = { left: 0, top: 0, right: DEFAULT_MARGIN, bottom: DEFAULT_MARGIN };` might make more sense, in case they've set `margin` to 0?  But also, that'll change the `vdiff`, so maybe not worth it lol...

svanherk (approved):

svanherk (approved):
🎉

bearfriend (approved):

PR #278, opened 2023-11-29
Add note about recursive waiting for getLoadingComplete

eKoopmans:
Wasn't sure the answer to this, thought it was worth clarifying in the readme.


dlockhart (approved):
Yep that's correct -- thanks!

PR #284, opened 2024-01-05
Fix Chrome rendering determinism issues

bearfriend:
When taking screenshots for vdiff tests in Chrome, the results could be non-deterministic depending on what kind of content was being rendered. In Unix systems, the content could sometimes render visually differently (though imperceptibly). Across platforms, content could sometimes render the same visually but something about the way it was rendered caused the screenshot bytes to change. Note that this was not an issue with the screenshot functionality or the PNG file encoding in any way. It was entirely caused by Chrome rendering the content differently between runs.

There was no perceptible change in performance with local testing.

Fixes:
- ~Byte size differences between vdiff test runs in Chrome~
- Non-flake pixel differences between vdiff test runs on Unix systems in Chrome

[GAUDI-370 Vdiff Chrome screenshot determinism](https://desire2learn.atlassian.net/browse/GAUD-370?atlOrigin=eyJpIjoiMjg1NDJkNjI3NzFkNDgxNTgwNzk2ODQ3OTNkOWMzYzkiLCJwIjoiaiJ9)


src/server/wtr-config.js
```diff
@@ -247,18 +247,19 @@ export class WTRConfig {
...
-			concurrency: b === 'firefox' || this.#cliArgs.open ? 1 : undefined, // focus in Firefox unreliable if concurrency > 1 (https://github.com/modernweb-dev/web/issues/238)
+			concurrency: b === BROWSER_MAP.firefox || this.#cliArgs.open ? 1 : undefined, // focus in Firefox unreliable if concurrency > 1 (https://github.com/modernweb-dev/web/issues/238)
 			product: b,
 			createBrowserContext: ({ browser }) => browser.newContext({ deviceScaleFactor: 2, reducedMotion: 'reduce' }),
 			launchOptions: {
 				headless: !this.#cliArgs.open,
 				devtools: false,
-				slowMo: this.#cliArgs.slowmo || 0
+				slowMo: this.#cliArgs.slowmo || 0,
+				args: b === BROWSER_MAP.chrome ? ['--disable-gpu-rasterization'] : []
```
bearfriend, line 262: `--disable-gpu-rasterization` seems to work just as well as `--disable-gpu` and `--disable-partial-raster` together, and seems better than disabling all hardware acceleration.

svanherk (approved):

bearfriend:
Oh sorry, that should have said "some" or "most" byte size differences 😆

PR #285, opened 2024-01-05
Gaud-370 - Detect ALL byte changes

svanherk:
To detect if a golden has the same pixels but metadata changes has caused the bytes to change, we currently try comparing the byte size between the golden and the new screenshot. I figured this was "good enough", because the chances the byte size would be identical but the bytes would be different would be very low. But turns out this is not that rare - the latest Chrome 120 changes caused 133 files to be in this situation (byte change with no byte size change).

So now, we actually do a buffer compare of the bytes to detect if something changed. I've found the perf on this to be not nearly as bad as I would expect - in `core`, it seems to add around 0ms to 100ms to each test, usually around 20ms to 40ms. And to help with this, I've moved the buffer compare up to earlier in the workflow so we can skip `pixelmatch` altogether if the two PNGs are an exact buffer match.

The workflow looks like this:
```mermaid
flowchart
    A[Do the screenshot and the golden have the same width and height?] -->|Yes| B[Do they have the same byte size?];
    A -->|No| C{Tests Fail:\nRun the existing resize logic.};
    B -->|Yes| D[Do they have the exact same buffers?];
    B -->|No| E[Run pixelmatch];
    D -->|Yes| F{Tests pass!}
    D -->|No| E
    E --> |pixelDiff == 0| G{Tests Fail:\nImage diff is clean but the images\ndo not have the same bytes.}
    E --> |pixelDiff != 0| H{Tests Fail:\nImage does not match golden.\n# pixels are different.}
```

### DIFF
--- a/src/server/visual-diff-plugin.js
+++ b/src/server/visual-diff-plugin.js
@@ -197,7 +197,8 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 
 				const rootLength = join(rootDir, PATHS.VDIFF_ROOT).length + 1;
 
-				const screenshotImage = PNG.sync.read(await readFile(screenshotFileName));
+				const screenshotFileBuffer = await readFile(screenshotFileName);
+				const screenshotImage = PNG.sync.read(screenshotFileBuffer);
 				setTestInfo(session, payload.name, {
 					slowDuration: payload.slowDuration,
 					new: {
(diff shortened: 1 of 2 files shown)


src/server/visual-diff-plugin.js
```diff
@@ -227,6 +229,14 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 				});
 
 				if (screenshotImage.width === goldenImage.width && screenshotImage.height === goldenImage.height) {
+					const goldenSize = (await stat(goldenFileName)).size;
+					const screenshotSize = (await stat(screenshotFileName)).size;
+					if (goldenSize === screenshotSize && screenshotFileBuffer.equals(goldenFileBuffer)) {
```
svanherk, line 234: We still do the quick byte size comparison here, to potentially skip the full buffer compare.

bearfriend (approved):

PR #306, opened 2024-02-05
Add style linting

bearfriend:
As other repos update to use `@brightspace-ui/stylelint-config@1`, if they don't use `--ignore-path .gitignore` when they run `stylelint` they may get some errors from the `.vdiff` folder because we haven't implemented it here yet.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,9 @@
   "version": "1.13.2",
   "repository": "https://github.com/BrightspaceUI/testing.git",
   "scripts": {
-    "lint": "eslint . --ext .js",
+    "lint": "npm run lint:eslint && npm run lint:style",
+    "lint:eslint": "eslint . --ext .js",
+    "lint:style": "stylelint \"**/*.{js,html}\" --ignore-path .gitignore",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -21,11 +23,13 @@
   "author": "D2L Corporation",
   "license": "Apache-2.0",
   "devDependencies": {
+    "@brightspace-ui/stylelint-config": "^1.0.0",
```
dlockhart, line 26: Nit:
```suggestion
    "@brightspace-ui/stylelint-config": "^1",
```

package.json
```diff
@@ -21,11 +23,13 @@
...
     "chai": "^5",
     "eslint": "^8",
     "eslint-config-brightspace": "^1",
     "mocha": "^10",
-    "sinon": "^17"
+    "sinon": "^17",
+    "stylelint": "^16.2.1"
```
dlockhart, line 32: Nit:
```suggestion
    "stylelint": "^16"
```

dlockhart (approved):
Good call. This is all going to go through Rollup, so hopefully it doesn't mess things up such that they fail.

dbatiste (approved):

PR #310, opened 2024-02-07
Add github actions run attempt reporter and cache action

bearfriend:
[GAUD-7538](https://desire2learn.atlassian.net/browse/GAUD-7538): d2l-test-runner: Add attempt reports

- Generate d2l-test-runner failed test reports
- Wrap the default reporter to fix skipped tests (via `--grep`) reporting as failed
- Add a modified cache action to make the previous run attempt report available to subsequent attempts

[GAUD-7538]: https://desire2learn.atlassian.net/browse/GAUD-7538?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


bearfriend:
- [ ] Wrap the action, or hardcode the inputs
- [ ] Merged test script reports instead of overwriting the file
- [ ] Fix default reporter progress (not just the final result)

.gitignore
```diff
@@ -1,2 +1,4 @@
+.DS_Store
 node_modules/
 .vdiff/
+.d2l-test/
```
dlockhart, line 4: I guess every repo that uses this will need to add this ignore entry?

bearfriend: No, it'll only exist in CI on test jobs. I just forgot to stop it from always writing the file because I was testing locally. I'll do that now.

src/browser/setup.js
bearfriend, line 1: Maybe this should go back to `vdiff` and I can pull our the mocha setup

bearfriend: Although, it is just being always included anyway

bearfriend:
Closing until reprioritize

PR #329, opened 2024-03-04
Fix `waitUntil` usage examples

bearfriend:
@TristanDevos [tried following](https://d2l.slack.com/archives/C04BR06RN22/p1709326105597629) these docs and ended up wasting some time with his `options` not working before finding the correct `open-wc` docs.


dlockhart (approved):
Oops, thanks!

PR #335, opened 2024-03-11
More realistic sendKeysElem timing

bearfriend:
Real user-browser interaction can't focus an element and type in the same task. This mimics a more realistic user experience, allowing time for delayed tasks to complete.

In particular, Safari is [flaking](https://github.com/BrightspaceUI/core/actions/runs/8210101107/job/22457315479?pr=4505#step:8:146) on a test where focus is retargeted after a timeout.


dlockhart (approved):
👍

PR #336, opened 2024-03-11
Replace unsafe characters in screenshot file names

bearfriend:
Windows has a set a characters that are not allowed in file paths, so we should remove them before naming the screenshot files. Linux doesn't allow `/` but that's already included.

This approach replaces each character with a `-`, which leaves a small possibility of conflicts, but I think the suggestion would just be to be more specific in that case.


dlockhart (approved):
This will be somewhat "breaking" in that goldens will get renamed, but probably fine.

bearfriend:
> This will be somewhat "breaking" in that goldens will get renamed, but probably fine.

Yeah, hopefully since no one has brought this up it won't affect anyone yet

PR #340, opened 2024-03-14
GAUD-6130 - Allow to override deviceScaleFactor

svanherk:
This is needed to make smaller vdiffs for the daylight site. You'd set the override at the top level of the config, and it's applied to all groups that config controls. I did this because I don't think we can set anything for the `test` or `vdiff` groups we've configured.

I didn't add any documentation about this because I don't think we _want_ people messing with this - thoughts?

### DIFF
--- a/src/server/wtr-config.js
+++ b/src/server/wtr-config.js
@@ -231,7 +231,7 @@ export class WTRConfig {
 		}
 
 		// convert all browsers to playwright
-		config.groups.forEach(g => g.browsers = this.getBrowsers(g.browsers));
+		config.groups.forEach(g => g.browsers = this.getBrowsers(g.browsers, config.deviceScaleFactor ?? (g.name === 'vdiff' ? 2 : 1)));
 
 		if (open || watch) {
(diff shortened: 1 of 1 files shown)


dlockhart (approved):
Cool! Yeah I'm good leaving this as undocumented for now.

src/server/wtr-config.js
```diff
@@ -231,7 +231,7 @@ export class WTRConfig {
 		}
 
 		// convert all browsers to playwright
-		config.groups.forEach(g => g.browsers = this.getBrowsers(g.browsers));
+		config.groups.forEach(g => g.browsers = this.getBrowsers(g.browsers, config.deviceScaleFactor ?? (g.name === 'vdiff' ? 2 : 1)));
```
svanherk, line 234: If we don't want to mess with switching this back to 1 for everything non-`vdiff`, we could just do:
```
config.deviceScaleFactor ?? 2
```
or
```
g.name === 'vdiff' && config.deviceScaleFactor ? config.deviceScaleFactor : 2
```

bearfriend (approved):

PR #431, opened 2024-07-09
fix: split paths out into its own file

dlockhart:
This is kind of a strange thing to do, but it's the only way I can get `insights-in-ellucian-experience` to generate a vdiff report. That repo is in a bit of an odd mix of `.mjs` / `.js` files, which I think is causing Rollup when building the report to treat something as CommonJS when it otherwise wouldn't and blow up.

By making this change, `PATHS` can be imported into the Rollup build without including all the other vdiff things -- one of which is the culprit.


bearfriend (approved):
🙈

PR #437, opened 2024-07-15
feat: add in `d2l-test-reporting` custom reporter and option

devpow112:
Preliminary work for adding in the [`d2l-test-reporting`](https://www.npmjs.com/package/d2l-test-reporting) custom reporter to be able to output our custom format for test meta-data. You can enable test reporting by either adding `testReporting: true` to your `d2l-test-runner.config.js` or using a CLI option `d2l-test-runner axe --chrome --test-reporting`.

Tested this out with `core` and it seemed to work as expected. It generated a report per test run for the various types of tests.

### DIFF
--- a/src/server/cli/test-runner.js
+++ b/src/server/cli/test-runner.js
@@ -125,7 +125,7 @@ async function getTestRunnerOptions(argv = []) {
 			name: 'test-reporting',
 			type: Boolean,
 			defaultOption: false,
-			description: 'Enables generating a D2L Test Reporting report JSON file',
+			description: 'Enables generating a D2L Test Reporting format JSON file',
 			order: 14
 		}
 	];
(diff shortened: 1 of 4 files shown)


devpow112:
FYI @Mark-Weiss this integrated pretty easily. Just doing some more testing.

src/server/cli/test-runner.js
```diff
@@ -121,6 +121,13 @@ async function getTestRunnerOptions(argv = []) {
 			description: 'Run tests in Webkit',
 			order: 4
 		},
+		{
+			name: 'test-reporting',
+			type: Boolean,
+			defaultOption: false,
+			description: 'Enables generating a D2L Test Reporting format JSON file',
+			order: 14
+		}
```
dlockhart, line 130: Would it make sense to just always enable reporting and either not make it an option, or enable it by default?

devpow112: I was thinking of this but wanted to try out a couple of integrations first. Then we can have it default to always generate.

dlockhart (approved):

bearfriend:
I somehow missed all this. Exciting! 🎉

PR #445, opened 2024-07-24
feat: update test reporting default

devpow112:
Enable test reporting by default when `GITHUB_ACTIONS` environment variable defined

### DIFF
--- a/src/server/wtr-config.js
+++ b/src/server/wtr-config.js
@@ -1,4 +1,5 @@
 import { defaultReporter } from '@web/test-runner';
+import { env } from 'node:process';
 import { headedMode } from './headed-mode-plugin.js';
 import { playwrightLauncher } from '@web/test-runner-playwright';
 import { reporter as testReportingReporter } from 'd2l-test-reporting/reporters/web-test-runner.js';
@@ -6,6 +7,7 @@ import { visualDiff } from './visual-diff-plugin.js';
 import { visualDiffReporter } from './visual-diff-reporter.js';
 
 const DEFAULT_PATTERN = type => `./test/**/*.${type}.js`;
+const DEFAULT_TEST_REPORTING = !!env['GITHUB_ACTIONS'];
 const BROWSER_MAP = {
 	chrome: 'chromium',
 	chromium: 'chromium',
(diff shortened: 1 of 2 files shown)


src/server/wtr-config.js
```diff
@@ -215,8 +217,6 @@ export class WTRConfig {
 			});
 		}
 
-		testReporting = !!testReporting || this.#cliArgs['test-reporting'];
```
dlockhart, line 218: What's the end result going to be for a random repo that didn't previously have this configured? Like there'll now be a JSON file that gets generated when the tests run? Are we worried that it won't be in `.gitignore`?

devpow112: Yeah there will be a `d2l-test-report.json` generated. I don't think it will be a problem for repos. Most repos don't explicitly check in generated files or do anything after running tests.

devpow112: I could add a disable test reporting flag if we think we will need to disable it.

src/server/wtr-config.js
```diff
@@ -1,11 +1,13 @@
 import { defaultReporter } from '@web/test-runner';
+import { env } from 'node:process';
 import { headedMode } from './headed-mode-plugin.js';
 import { playwrightLauncher } from '@web/test-runner-playwright';
 import { reporter as testReportingReporter } from 'd2l-test-reporting/reporters/web-test-runner.js';
 import { visualDiff } from './visual-diff-plugin.js';
 import { visualDiffReporter } from './visual-diff-reporter.js';
 
 const DEFAULT_PATTERN = type => `./test/**/*.${type}.js`;
+const DEFAULT_TEST_REPORTING = !!env['GITHUB_ACTIONS'];
```
devpow112, line 10: I had changed this to `GITHUB_ACTIONS` but wonderig if I should leave it as `CI` since this is more generic. Thoughts?

src/server/wtr-config.js
dlockhart: Ok cool. I don't think it's worth adding a flag until we know we'll need it.

dlockhart (approved):

dlockhart (approved):

PR #447, opened 2024-07-25
feat: remove references to old visual-diff library and migration tools

dlockhart:
All repos have been migrated to the new library, so this removes references to the migration guide & tools.


bearfriend (approved):
Hell yeah

PR #462, opened 2024-08-13
fix: split byte diff filtering out from test status

dlockhart:
Treating tests that have failed but only because of file size differences as a "status" ended up being problematic with multiple browsers. A test in one browser might have a byte diff failure whereas the other failed for another reason, so the counts get all messed up.

This changes things around to treat it as its own filter on top of whatever status filter is applied, which works out much nicer.

Screenshot 2024-08-13 at 4 07 26 PM

### DIFF
--- a/src/server/report/app.js
+++ b/src/server/report/app.js
@@ -306,7 +306,7 @@ class App extends LitElement {
 			if (!this._filterBrowsers.includes(b.name)) return nothing;
 			const result = test.results.find(r => r.name === b.name);
 			const passed = (result !== undefined) ? result.passed : true;
-			const text = passed ? 'passed' : result.bytediff ? 'byte-diff' : 'failed';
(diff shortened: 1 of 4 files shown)


src/server/report/app.js
```diff
@@ -289,7 +306,7 @@ class App extends LitElement {
 			if (!this._filterBrowsers.includes(b.name)) return nothing;
 			const result = test.results.find(r => r.name === b.name);
 			const passed = (result !== undefined) ? result.passed : true;
-			const text = passed ? 'passed' : 'failed';
+			const text = passed ? 'passed' : result.bytediff ? 'byte-diff' : 'failed';
```
bearfriend, line 309: nit:
```suggestion
			const text = passed ? 'passed' : (result.bytediff ? 'byte-diff' : 'failed');
```

bearfriend (approved):

PR #466, opened 2024-08-16
feat: set isD2LTestPage flag

dlockhart:
Sets a flag to indicate that we're in a test environment. This will allow places [like feature flags](https://github.com/BrightspaceUI/lms-core/pull/160) to operate slightly differently.


bearfriend (approved):

bearfriend:
There was something recently that we checked a URL for because we didn't have this. Maybe one of the weird vdiff tests that needed test data or something?

PR #472, opened 2024-08-22
remove duplicate docs

dlockhart:
Had these changes unsaved in my editor. This was a copy-paste from the above example where I initially forgot to remove the one code sample.


bearfriend (approved):
I almost mentioned that the placement seemed odd but didn't realize it was duplicate 😆

PR #527, opened 2024-10-25
Add system/os data

bearfriend:
Adds system data to track changes between runs. This should solve confusion around diffs where there were no code changes but any of the OS platform, OS release, or system architecture changes between runs, such as when our CI runners are updated, or a local repo is copied to a new machine.

Example:
Screenshot 2024-10-25 at 5 30 11 PM


dlockhart (approved):
Nice!

PR #542, opened 2024-11-13
feat: add `clickElemAt` & `hoverElemAt`

### LINKED ISSUE #281: [Feature-Request] `clickElemAt` + `hoverElemAt`
### Problem Definition
- Many components have specific behaviours depending on which portion of the element is being interacted with
- Specifically in some cases, a child component may rest in the very center of the parent component that is being interacted with
  - This child component will now be the main target of `clickElem` or `hoverElem` and can potentially hijack the expected experience of the target element for its own

### Current workarounds
  - dig through the shadowRoot of the parent to find a different child element that won't hijack the expected behaviour from parent (if one exists)
  - home brew a `clickAt` or `hoverAt` solve by using `elem.getBoundingClientRect()` to generate a `DOMRect()` for position target information

### Suggestion
- Provide new functions `clickElemAt` and `hoverElemAt` which will interact with the target element at some offset from their origin instead
- [Example branch](https://github.com/BrightspaceUI/testing/compare/main...tkearse/feature-request-clickAt-hoverAt)

KearseTrevor:
Resolves #281 

Alternative setup would be `clickElemAt(elem, xOffset = 0, yOffset = 0)` which likely brings it in line closer to how users expect things to work when coming from `clickAt` and `hoverAt` but I have it penciled in as `clickElemAt(elem, offset = { x: 0, y: 0 })` right now just because it felt cleaner to me but it's a quick change.

Honestly on second thought, I'll likely just swap to using it as xOffset, yOffset since it makes more sense to follow pattern.

Just adding documentation and then it's ready for full review


test/browser/commands.test.js
```diff
@@ -31,6 +31,38 @@ describe('commands', () => {
 			window.removeEventListener('click', onClick);
 		});
 
+		it('should click at top-left by default', async() => {
+			const clickPos = { x: 0, y: 0 };
```
dlockhart, line 35: Since this and the `mousePos` stuff is now used in a bunch of tests, can we pull it out and set it up in a `before` just inside the `describe('click/hover')` block?

src/browser/commands.js
```diff
@@ -5,6 +5,8 @@ export { setViewport } from './reset.js';
 function getElementPosition(elem) {
 	const { x, y, width, height } = elem.getBoundingClientRect();
 	return {
+		left: Math.floor(x + window.scrollX),
+		top: Math.floor(y + window.scrollY),
```
dlockhart, line 9: The only thing that might be an issue here is right-to-left, where the left/right would need to flip. Would it make any sense for the offsets to be from centre instead of top-left? 🤔

test/browser/commands.test.js
KearseTrevor: Yup sounds good! Thanks for the rec!

src/browser/commands.js
dlockhart: Very true... and yeah the existing stuff clicking in the centre is already susceptible to this like you said. Let's ignore this for now -- if it comes up we can address it (or not) then.

test/browser/commands.test.js
```diff
@@ -19,18 +31,35 @@ describe('commands', () => {
 		});
 
 		it('should click at position', async() => {
-			const clickPos = { x: 0, y: 0 };
-			function onClick(e) {
-				clickPos.x = e.clientX;
-				clickPos.y = e.clientY;
-			}
 			window.addEventListener('click', onClick);
```
dlockhart, line 34: Let's add the event listeners in a global `before` (just runs once) and remove them in a global `after`.

test/browser/commands.test.js
```diff
@@ -115,29 +165,33 @@ describe('commands', () => {
...
 		const mousePos = { x: 0, y: 0 };
 		function onMouseMove(e) {
 			mousePos.x = e.clientX;
 			mousePos.y = e.clientY;
 		}
 
-		const buttonTemplate = html`<button>text</button>`;
+		before(() => {
+			window.addEventListener('mousemove', onMouseMove);
+		});
```
dlockhart, line 178: Didn't see this stuff doing the same thing down here -- could just move it up under the `commands` fixture and then every test in the whole file can use it.

dlockhart (approved):
Awesome, thanks for the refactor of the tests!

PR #554, opened 2024-11-27
MAN-3694 -- Add a waitUntilSettled function that waits for an element to finish updating

mpharoah-d2l:
https://desire2learn.atlassian.net/browse/MAN-3694

This PR adds a public `waitUntilSettled` function that more reliably waits for an element to finish updating. Our team intends to use this in multiple repos to help reduce vdiff flake.

### DIFF
--- a/src/browser/wait.js
+++ b/src/browser/wait.js
@@ -1,0 +1,173 @@
+import { aTimeout, nextFrame } from '@open-wc/testing';
+
+class DomSnapshot {
+	constructor(element) {
+		this._roots = [element.outerHTML];
+		this._addShadowRoots = this._addShadowRoots.bind(this);
+		this._addShadowRoots(element);
+	}
+
+	equals(other) {
+		if (!other || this._roots.length !== other._roots.length) {
+			return false;
+		}
+
+		for (let i = 0; i < this._roots.length; i++) {
+			if (this._roots[i] !== other._roots[i]) {
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	_addShadowRoots(element) {
+		if (element.shadowRoot) {
+			this._roots.push(element.shadowRoot.innerHTML);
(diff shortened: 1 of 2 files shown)


src/browser/wait.js
```diff
@@ -0,0 +1,173 @@
...
+	}
+};
+
+async function _waitUntilSettledImpl(element, options) {
+	const hasSirenActionQueue = options.awaitHypermedia &&
+		window.D2L &&
+		D2L.Siren &&
+		D2L.Siren.ActionQueue &&
+		D2L.Siren.ActionQueue.isPending &&
+		D2L.Siren.ActionQueue.enqueue;
```
dlockhart, line 109: This feels like too much D2L-specific knowledge leakage into what we're hoping to keep as a bare bones testing library. Have you looked at the [getLoadingComplete stuff](https://github.com/BrightspaceUI/testing?tab=readme-ov-file#waiting-for-asynchronous-components) ([or mixin](https://github.com/BrightspaceUI/core/tree/main/mixins/loading-complete)) to handle this kind of thing? Each component that does async calls would need to be aware of it, but that's the way we'd always imagined this stuff working.

If you did want to try and write a single thing that "waits for things to settle", you could still use `getLoadingComplete` in a shared mixin and put this logic there.

Alternately, you could create your own `fixture` method that calls into `@brightspace-ui/testing`'s `fixture` (which calls into the `open-wc` one) but then additionally waits for more stuff?

dlockhart: Hmm I'm not sure... there's a lot of Siren-specific stuff in here so maybe as a testing utility in `siren-sdk`? Or does your team have a shared utils library?

mpharoah-d2l: No, but I guess we can just make one

mpharoah-d2l:
Going to just copy paste it in d2l-outcomes instead

PR #567, opened 2024-12-11
Make report test names copy-friendly

bearfriend:
Found this sitting around. It makes the test names in the report copyable for using with the `grep` option.


dlockhart (approved):
Nice!

PR #646, opened 2025-03-27
GAUD-7681: treat byte diffs properly

dlockhart:
When working in core on a change that caused a large number of zero pixel but many byte diffs, I realized that the vdiff report was misidentifying them.

### DIFF
--- a/src/server/visual-diff-reporter.js
+++ b/src/server/visual-diff-reporter.js
@@ -101,7 +101,7 @@ function flattenResults(session, browserData, fileData) {
 				});
 			}
 			const testData = fileData.tests.get(testName);
-			const bytediff = !t.passed && !!info?.diff && info?.pixelsDiff === 0;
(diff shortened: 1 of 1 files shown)


GZolla (approved):

src/server/visual-diff-reporter.js
```diff
@@ -101,7 +101,7 @@ function flattenResults(session, browserData, fileData) {
 				});
 			}
 			const testData = fileData.tests.get(testName);
-			const bytediff = !t.passed && !!info?.diff && info?.pixelsDiff === 0;
+			const bytediff = !t.passed && info?.diff === undefined && info?.pixelsDiff === 0;
```
dlockhart, line 104: `info.diff` is going to be set to a string value of the path to the diff PNG. When `pixelsDiff` is `0` -- i.e. when there's a byte diff only, then `info.diff` will be `undefined`. `!!someString` will always yield `true` and `!!undefined` will yield `false`, which is the opposite of what we want.

bearfriend (approved):

PR #665, opened 2025-04-22
GAUD-7838 - Force golden mode

svanherk:
Testing if the element-different goldens are up-to-date. ~May not have to merge this.~

Ok yeah, we do need to merge this. These goldens weren't showing up as a problem in the other PR because the tests check that the element is different and fail (as expected). But we want to make sure the element is different because of the things we're testing, not because of these aliasing differences.

Will be creating a story to handle this better.


svanherk:
Ok, now going to put it back into regular mode and merge this into https://github.com/BrightspaceUI/testing/pull/663

bearfriend (approved):

PR #730, opened 2025-07-14
GAUD-8264 - Rewrite the element-different tests to not commit goldens

svanherk:
Currently, the goldens for the expected-to-fail `element-different` tests can get out of date, because they'll still properly fail so we don't notice. And when a vdiff PR is put up, that PR is wrong and we need to re-run it with a temporary change we then uncommit. This doesn't happen often and is confusing to investigate each time.

Instead of all that, we'll no longer commit the goldens for the `element-different` tests. Instead, we just grab the default golden,  copy it in before the test, confirm it fails, then remove it before the vdiff action copies the passed and failed screenshots.

The commits in here are a mess (I'll be squashing them) because of all my testing, but [here's](https://vdiff.d2l.dev/BrightspaceUI/testing/5a49398a1af29e9f0d690bdd1f1b2d07ec01ec88/2025-07-14_18-56-31/report/) what a report looks like if an `element-different` test fails (meaning the two images were actually the same, which is not expected). We wouldn't merge that PR in, but this should be obvious that something is wrong.


test/browser/element.vdiff.js
```diff
@@ -179,6 +181,7 @@ describe('element-matches', () => {
 
 });
 
+const expectFailures = true; // Easy switch to create traditional report for debugging
```
svanherk, line 184: Added this as an easy way to debug locally - flipping this value makes the tests run "normally" and all fail, resulting in a report with the diffs.

test/browser/element.vdiff.js
```diff
@@ -223,29 +232,31 @@ describe('element-different', () => {
 				fail = true;
 			}
 
-			expect(fail, 'current and golden images to be different').equal(!isGolden);
-
-			if (!isGolden) {
-				await executeServerCommand('vdiff-revert-golden-file', { testCategory: 'element-different', fileName: `${name}.png` });
-			}
+			expect(fail, 'current and golden images to be different').equal(expectFailures);
+			await executeServerCommand('vdiff-remove-test-files', { testCategory: 'element-different', fileName: `${name}.png` });
```
svanherk, line 236: We run this _after_ the `expect` so that if it fails (i.e. the images are unexpectedly the same), we don't clean up the files so a commit can happen and the report will be uploaded and use them.

svanherk:
I'm not deleting the orphaned goldens here because the next run will clean them up, but I think I want that run to be a clean test so I'll add them.

PR #740, opened 2025-07-24
Additional CSS variables for RTL.

dbatiste:
[GAUD-8448](https://desire2learn.atlassian.net/browse/GAUD-8448)

This PR adds a couple more CSS variables that are needed to address some RTL cases that CSS logical properties does not support.

[GAUD-8448]: https://desire2learn.atlassian.net/browse/GAUD-8448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/wtr-config.js
+++ b/src/server/wtr-config.js
@@ -82,16 +82,22 @@ export class WTRConfig {
 						<style>
 
 							html {
+								--d2l-cursor-resize-inline-end: e-resize;
+								--d2l-cursor-resize-inline-start: w-resize;
 								--d2l-document-direction: ltr;
 								--d2l-inline-end: right;
 								--d2l-inline-start: left;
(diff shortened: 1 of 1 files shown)


src/server/wtr-config.js
```diff
@@ -82,16 +82,22 @@ export class WTRConfig {
 						<style>
 
 							html {
+								--d2l-cursor-resize-inline-end: e-resize;
```
dbatiste, line 85: Hopefully the resize ones are obvious.

The `--d2l-length-factor` is a little more general, and can be used to invert a dimension where CSS logical properties do not apply (for example, properties of `box-shadow`).

PR #947, opened 2026-04-07
feat: add alt tests

GZolla:
[GAUD-9495](https://desire2learn.atlassian.net/browse/GAUD-9495)

Quite a big refactor but essentially this enables a `dark` option on `golden` that will add a separate test for dark.

[GAUD-9495]: https://desire2learn.atlassian.net/browse/GAUD-9495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/visual-diff-plugin.js
+++ b/src/server/visual-diff-plugin.js
@@ -20,14 +20,6 @@ const ALT_TESTS = {
 	}
 };
 
-const ERROR_MESSAGES = {
-	noGolden: 'No golden exists. Use the "--golden" CLI flag to re-run and re-generate goldens.',
-	moveFile: 'Problem moving file to "pass" directory.',
-	pixelDiff: (pixelsDiff) => `Image does not match golden. ${pixelsDiff} pixels are different.`,
-	byteDiff: 'Image diff is clean but the images do not have the same bytes.',
-	resizeDiff: (bestPixelsDiff) => `Images are not the same size. When resized, ${bestPixelsDiff} pixels are different.`
-};
-
 async function checkFileExists(fileName) {
 	try {
 		await access(fileName, constants.F_OK);
@@ -84,6 +76,45 @@ function getResizedPng(xName, yName, original, newSize) {
 	const resized = new PNG(newSize);
 	PNG.bitblt(original, resized, 0, 0, original.width, original.height, xOffsets[xName], yOffsets[yName]);
 	return resized;
+}
+
+let bestPosition = null;
+async function getBestResizePng(screenshotImage, goldenImage) {
+	const newWidth = Math.max(screenshotImage.width, goldenImage.width);
+	const newHeight = Math.max(screenshotImage.height, goldenImage.height);
+	const newSize = { width: newWidth, height: newHeight };
+
+	if (bestPosition !== null) {
+		const [y, x] = bestPosition.split('-');
+		const screenshot = getResizedPng(x, y, screenshotImage, newSize);
+		const golden = getResizedPng(x, y, goldenImage, newSize);
+		const { diff, pixelsDiff } = getPixelsDiff(screenshot, golden);
+		return { screenshot, golden, diff, pixelsDiff };
+	}
+
+	const newScreenshots = await createComparisonPNGs(screenshotImage, newSize);
+	const newGoldens = await createComparisonPNGs(goldenImage, newSize);
+
+	let bestIndex = -1;
+	let bestDiffImage = null;
+	let bestPixelsDiff = Number.MAX_SAFE_INTEGER;
+	for (let i = 0; i < newScreenshots.length; i++) {
+		const { diff, pixelsDiff } = getPixelsDiff(newScreenshots[i].png, newGoldens[i].png);
+
+		if (pixelsDiff < bestPixelsDiff) {
+			bestIndex = i;
+			bestDiffImage = diff;
+			bestPixelsDiff = pixelsDiff;
+		}
+	}
+
+	bestPosition = newScreenshots[bestIndex].position;
+	return {
+		screenshot: newScreenshots[bestIndex].png,
+		golden: newGoldens[bestIndex].png,
+		diff: bestDiffImage,
+		pixelsDiff: bestPixelsDiff
+	};
 }
 
 async function createComparisonPNGs(original, newSize) {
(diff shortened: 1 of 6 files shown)


src/server/visual-diff-plugin.js
```diff
@@ -145,10 +164,14 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
 			const testPath = dirname(session.testFile).replace(rootDir, '');
 			const newPath = join(rootDir, PATHS.VDIFF_ROOT, testPath, dir);
-			const goldenFileName = `${join(newPath, PATHS.GOLDEN, browser, newName)}.png`;
-			const passFileName = `${join(newPath, PATHS.PASS, browser, newName)}.png`;
+			const goldenFile = join(newPath, PATHS.GOLDEN, browser, newName);
+			const passFile = join(newPath, PATHS.PASS, browser, newName);
 			const screenshotFile = join(newPath, PATHS.FAIL, browser, newName);
-			const screenshotFileName = `${screenshotFile}.png`;
+			function getFileName(base, alt = 'default') {
+				return `${base}${alt === 'default' ? '' : `-${alt}`}.png`;
```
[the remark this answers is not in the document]
dbatiste: The naming is really inconsistent in the core components. In some cases we prefix, in other cases we suffix. In the dark mode tests so far, I've been trying to be consistent within the given test file, knowing that tests will likely eventually be reorganized, with some names changing.

src/server/visual-diff-plugin.js
```diff
@@ -244,17 +280,45 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 									byteSize: screenshotSize
 								},
 								pixelsDiff
-							});
-							return { pass: false, message: 'Image diff is clean but the images do not have the same bytes.' };
+							}, alt);
+							return { pass: false, message: ERROR_MESSAGES.byteDiff };
 						}
 					} else {
 						return { resizeRequired: true };
```
svanherk: Any more thoughts on my comment [here](https://github.com/BrightspaceUI/testing/pull/939#discussion_r3018676838)? I do think it'd make this a bit cleaner - all the resize command code would move in here, and then you don't need to do special handling and re-setting variables and such. But it's not necessarily risk-free.

src/server/visual-diff-plugin.js
```diff
@@ -210,19 +242,23 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 							path: goldenFileName.substring(rootLength),
 							width: goldenImage.width
 						}
-					});
+					}, alt);
+
+					if (resizeRequired) {
+						return { resizeRequired: true }; // If any test requires a resize, assume alt tests will too
```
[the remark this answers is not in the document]
GZolla: Now it does! The refactor solved more than one issue, now it only runs resize logic if it actually needs to resize, with the best position being remembered if any test resized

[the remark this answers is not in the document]
svanherk: Awesome! 🎉

src/server/visual-diff-plugin.js
```diff
@@ -261,7 +292,7 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
 						infoManager.set({
 							diff: `${screenshotFile.substring(rootLength)}-diff.png`,
```
svanherk: Does this look like `button-primary.dark-diff.png`? Or like `button-primary-diff.png`? If it's the first, that looks a little weird. If it's the second, I think we need to add the `alt` in here as well.

src/server/visual-diff-plugin.js
```diff
@@ -261,7 +292,7 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
 							pixelsDiff
-						});
+						}, alt);
 						await writeFile(`${screenshotFile}-diff.png`, PNG.sync.write(diff));
```
svanherk: Similar question here, what do the diff file names look like?

src/server/visual-diff-plugin.js
```diff
@@ -273,30 +304,47 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
 					}
 				} else {
 					const { screenshot, golden, diff, pixelsDiff } = await getBestResizePng(screenshotImage, goldenImage);
 
-					await writeFile(`${screenshotFile}-resized-screenshot.png`, PNG.sync.write(screenshot.png));
-					await writeFile(`${screenshotFile}-resized-golden.png`, PNG.sync.write(golden.png));
-					await writeFile(`${screenshotFile}-diff.png`, PNG.sync.write(diff));
+					const screenshotResizedPath = getFileName(`${screenshotFile}-resized`, alt);
+					const goldenResizedPath = getFileName(`${goldenFile}-resized`, alt);
+					const diffPath = getFileName(`${screenshotFile}-diff`, alt);
```
svanherk: These seem correct, so I think it's just the "same size" path above that needs to be updated to match.

src/server/visual-diff-plugin.js
```diff
@@ -273,30 +304,47 @@ export function visualDiff({ updateGoldens = false, runSubset = false } = {}) {
...
 						return { pass: false, message: 'Image diff is clean but the images do not have the same bytes.' };
 					}
 				} else {
 					const { screenshot, golden, diff, pixelsDiff } = await getBestResizePng(screenshotImage, goldenImage);
 
-					await writeFile(`${screenshotFile}-resized-screenshot.png`, PNG.sync.write(screenshot.png));
-					await writeFile(`${screenshotFile}-resized-golden.png`, PNG.sync.write(golden.png));
-					await writeFile(`${screenshotFile}-diff.png`, PNG.sync.write(diff));
+					const screenshotResizedPath = getFileName(`${screenshotFile}-resized`, alt);
+					const goldenResizedPath = getFileName(`${goldenFile}-resized`, alt);
```
svanherk: Actually, I think these two are slightly off. We want them both in the failed folder, identifying what type they are. Something like:
```suggestion
					const screenshotResizedPath = getFileName(`${screenshotFile}-screenshot-resized`, alt);
					const goldenResizedPath = getFileName(`${screenshotFile}-golden-resized`, alt);
```
if we want to match the old names (while including the new `.<alt>.png` piece).

src/server/visual-diff-reporter.js
```diff
@@ -85,39 +85,43 @@ function flattenResults(session, browserData, fileData) {
...
+				}
+				const testData = fileData.tests.get(testName);
+				const bytediff = !t.passed && info?.diff === undefined && info?.pixelsDiff === 0;
+				if (!t.passed) {
+					if (bytediff) {
+						browserData.numByteDiff++;
+						testData.numByteDiff++;
+					}
+					browserData.numFailed++;
+					testData.numFailed++;
```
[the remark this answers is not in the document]
svanherk: Chatted offline - we're good to leave this for now and we can rethink/revisit with the report redesign. You can still get the actual "test number" in the logs.

svanherk (approved):

PR #962, opened 2026-04-20
fix: mark just failed alt-tests as failed

GZolla:
[GAUD-9495](https://desire2learn.atlassian.net/browse/GAUD-9495)

If alt tests are defined and any of them fail, the reporter would mark(and count) ALL the alt tests as failed, even if they passed

[GAUD-9495]: https://desire2learn.atlassian.net/browse/GAUD-9495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/server/visual-diff-info.js
+++ b/src/server/visual-diff-info.js
@@ -20,6 +20,7 @@ export class TestInfoManager {
 				testInfo.new = { ...info.new, ...testInfo.new };
 			}
 			testInfo.diff = testInfo.diff || info.diff;
+			testInfo.byteSizeDiff = testInfo.byteSizeDiff || info.byteSizeDiff;
 		}
 		const tests = testInfoMap.get(this.key) || {};
 		testInfoMap.set(this.key, { ...tests, [alt]: testInfo });
--- a/src/server/visual-diff-reporter.js
+++ b/src/server/visual-diff-reporter.js
@@ -106,7 +106,7 @@ function flattenResults(session, browserData, fileData) {
 				const testData = fileData.tests.get(testName);
 				const passed = !!(info?.golden) && !(info?.diff) && !(info?.pixelsDiff);
 				const bytediff = info?.diff === undefined && info?.pixelsDiff === 0;
-				if (passed) {
+				if (!passed) {
 					if (bytediff) {
 						browserData.numByteDiff++;
 						testData.numByteDiff++;
--- a/src/server/visual-diff-reporter.js
+++ b/src/server/visual-diff-reporter.js
@@ -104,10 +104,9 @@ function flattenResults(session, browserData, fileData) {
 					});
 				}
 				const testData = fileData.tests.get(testName);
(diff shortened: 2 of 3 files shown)


src/server/visual-diff-reporter.js
```diff
@@ -104,8 +104,9 @@ function flattenResults(session, browserData, fileData) {
 					});
 				}
 				const testData = fileData.tests.get(testName);
-				const bytediff = !t.passed && info?.diff === undefined && info?.pixelsDiff === 0;
-				if (!t.passed) {
+				const passed = !!(info?.golden) && !(info?.diff) && !(info?.pixelsDiff);
+				const bytediff = info?.diff === undefined && info?.pixelsDiff === 0;
+				if (passed) {
```
svanherk: Should this be `!passed`?

src/server/visual-diff-reporter.js
```diff
@@ -104,8 +104,9 @@ function flattenResults(session, browserData, fileData) {
 					});
 				}
 				const testData = fileData.tests.get(testName);
-				const bytediff = !t.passed && info?.diff === undefined && info?.pixelsDiff === 0;
-				if (!t.passed) {
+				const passed = !!(info?.golden) && !(info?.diff) && !(info?.pixelsDiff);
+				const bytediff = info?.diff === undefined && info?.pixelsDiff === 0;
```
svanherk: Hmm, still not sure what to do about this - how do we tell if an individual test passed or if it failed but there were only byte differences? In both cases, there's a golden, no diff, and pixelsDiff is 0.

svanherk: Maybe we should just add it to the TestInfo when the test is running, instead of trying to infer?

svanherk: Or at least maybe a new property like `byteDiffFailure` that's set [here](https://github.com/BrightspaceUI/testing/blob/0e614bd030c4a7976687720f65bf12fb26357651/src/server/visual-diff-plugin.js#L299-L307), and then these check:
```
const passed = !!(info?.golden) && !(info?.diff) && !(info?.pixelsDiff) && !(info?.byteDiffFailure);
if (!passed) {
	if (info?.byteDiffFailure) {
		...
```

src/server/visual-diff-info.js
```diff
@@ -20,6 +20,7 @@ export class TestInfoManager {
 				testInfo.new = { ...info.new, ...testInfo.new };
 			}
 			testInfo.diff = testInfo.diff || info.diff;
+			testInfo.byteSizeDiff = testInfo.byteSizeDiff || info.byteSizeDiff;
```
svanherk, line 23: Not really related, but just noticed - should `pixelsDiff` be here too? Maybe it's working fine because the manager is never called again after a call setting it?

svanherk (approved):
Other than the remaining question about `pixelsDiff`, this makes sense to me!

PR #993, opened 2026-05-21
GAUD-10062: expose waitForElem

dlockhart:
In somewhat rare circumstances, changing a component's state can cause a totally different set of components within it to render. One example might be putting it into an error state. When this happens, awaiting `updateComplete` isn't often enough, as those *new* nested components might have language terms that need to load or they may need to do their own asynchronous work.

To solve this, we expose `waitForElem` -- which is already used internally by the original `fixture` call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,9 +139,9 @@ it('should work when printing', async() => {
 });
 ```
 
-### Accessibility Testing with aXe
+### Accessibility Testing with axe
 
-Elements can be processed by the [aXe accessibility validator](https://github.com/dequelabs/axe-core), which will automatically fail the test if any violations are detected.
+Elements can be processed by the [axe accessibility validator](https://github.com/dequelabs/axe-core), which will automatically fail the test if any violations are detected.
 
 ```javascript
 it('should be accessible', async() => {
@@ -299,6 +299,20 @@ it('should wait for updates', async() => {
   const elem = await fixture(html`<my-elem></my-elem>`);
   elem.someProp = 'foo';
   await elem.updateComplete;
+});
+```
+
+If that change causes a different set of nested components to be rendered, and those nested components take some time to finish *their* rendering, it's possible that awaiting `updateComplete` isn't enough.
+
+In this case, `waitForElem` can be used to recursively wait in a similar way to the original `fixture` call:
+
+```javascript
+import { fixture, waitForElem } from '@brightspace-ui/testing';
+
+it('should wait for recursive re-rendering', async() => {
+  const elem = await fixture(html`<my-elem></my-elem>`);
+  elem.errorState = true;
+  await waitForElem(elem);
 });
 ```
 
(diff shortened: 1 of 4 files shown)


src/browser/fixture.js
```diff
@@ -36,20 +36,20 @@ function getComposedChildren(node) {
...
 
 	if (!elem) return;
 
 	const doWait = async() => {
 
-		const update = elem.updateComplete;
-		if (typeof update === 'object' && Promise.resolve(update) === update) {
-			await update;
+		if (awaitLoadingComplete && typeof elem.getLoadingComplete === 'function') {
+			await elem.getLoadingComplete();
```
dlockhart, line 46: I've also flipped the order here. We first wait for `getLoadingComplete` and then we wait for `updateComplete`. The reasoning here is that almost always `getLoadingComplete` resolving is going to trigger some kind of state change, so we wait for `updateComplete` after that.

README.md
```diff
@@ -139,9 +139,9 @@ it('should work when printing', async() => {
 });
 ```
 
-### Accessibility Testing with aXe
+### Accessibility Testing with axe
```
dlockhart, line 142: While I was here...

GZolla (approved):

test/browser/fixture.test.js
```diff
@@ -332,6 +350,17 @@ describe('fixture', () => {
...
 
+		it('should wait for elements that are updated to be slow', async() => {
+			const elem = await fixture(`<${fastOrSlowElem}></${fastOrSlowElem}>`);
+			expect(elem.shadowRoot.querySelector('p').textContent).to.equal('fast');
+			elem.slow = true;
+			const waitPromise = waitForElem(elem);
+			await waitUntil(() => resolves.has('slow'));
+			timeouts.push(setTimeout(() => resolves.get('slow')(), 50));
+			await waitPromise;
+			expect(elem.shadowRoot.querySelector(slowElem).finished).to.be.true;
```
bearfriend, line 362: Should we be concerned that the `waitUntil` is actually the thing delaying long enough to make the `expect` pass? Can we use `hasChanged` or an event instead? Does setting `slow` on the original fixture not hold `waitForElem` until everything is done?

```suggestion
			const waitPromise = waitForElem(elem);
			elem.slow = true;
			await waitPromise;
			expect(elem.shadowRoot.querySelector(slowElem).finished).to.be.true;
```

Image

Like `waitPromise` is never going to resolve before `elem.slow` is set, and then that would kick off the update cycle and should hold the `await` until `finished` is set... I think.

dlockhart: We discussed this offline, but the conclusion was: the nested slow element won't resolve its `getLoadingComplete` until we call `resolves.get('slow')()`, which happens after the `waitUntil`. So that part is safe. But just for extra safety, I added an assertion that `finished` is `false` right before we await `waitPromise`.

bearfriend (approved):

PR #995, opened 2026-05-22
GAUD-10062: fix webkit tests

dlockhart:
These tests only fail on my dev machine, not in CI. The calls to `getBoundingClientRect` return decimal values whereas the mouse positions are integers.


bearfriend (approved):
I'm surprised `floor` works instead of `ceil`/`round` 🤷

dlockhart:
> I'm surprised floor works instead of ceil 

Yeah they were all `.34375`.

PR #1005, opened 2026-06-05
GAUD-10114: upgrade and add test reporting

dlockhart:
v6 should be a drop-in replacement.

### DIFF
--- a/d2l-test-reporting.config.json
+++ b/d2l-test-reporting.config.json
@@ -1,0 +1,17 @@
+{
+  "tool": "UI Platform & Components",
+  "overrides": [
+    {
+      "pattern": "**/test/*.test.js",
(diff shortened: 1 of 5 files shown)


svanherk (approved):

d2l-test-reporting.config.json
svanherk, line 1: Is this because it wasn't categorizing our tests correctly?

dlockhart: The config file is required, I just missed it in the other repos.

On top of that, the `Error: Report path must exists` errors that I'm getting in `lms-core` is because `experience` is required in the config in v5 but dropped in v6 -- which is why this repo is working fine. I just confirmed with QE that releasing this is safe and should allow downstream repos to not set `experience`.

PR #1008, opened 2026-06-08
GAUD-10114: fix test categories

dlockhart:
These patterns weren't getting a hit since this repo is organized a bit differently, so they were falling until the "Other" category.


svanherk (approved):
We maybe could have done something to all of our repos to be more generic, like `**/test/**/*.test.js`, but 🤷‍♀️

